PR #28, opened 2023-09-13
Dashboard page

andre8244:
Add charts and cards to Dashboard page

### DIFF
--- a/src/components/standalone/dashboard/WanTrafficCard.vue
+++ b/src/components/standalone/dashboard/WanTrafficCard.vue
@@ -22,8 +22,8 @@ import { useI18n } from 'vue-i18n'
 const { t } = useI18n()
 const loginStore = useLoginStore()
 const REFRESH_INTERVAL = 10000
-const CYAN_300 = '#67e8f9'
-const CYAN_600 = '#0891b2'
+const CYAN_400 = '#22d3ee'
+const GRAY_500 = '#6b7280'
 const chartLabels = ref<any[]>([])
 const chartDatasets = ref<any[]>([])
 const intervalId = ref<number>(0)
@@ -125,16 +125,16 @@ async function getInterfaceTraffic() {
     chartDatasets.value = [
       {
         label: t('standalone.dashboard.download'),
-        borderColor: CYAN_300,
-        backgroundColor: CYAN_300,
+        borderColor: CYAN_400,
+        backgroundColor: CYAN_400,
         data: downloadData,
         borderWidth: 1,
         radius: 0
       },
       {
         label: t('standalone.dashboard.upload'),
-        borderColor: CYAN_600,
-        backgroundColor: CYAN_600,
+        borderColor: GRAY_500,
+        backgroundColor: GRAY_500,
         data: uploadData,
         borderWidth: 1,
         radius: 0


src/components/NeCard.vue
```diff
@@ -0,0 +1,81 @@
...
+-->
+
+<script setup lang="ts">
+import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome'
+import { NeSkeleton, NeInlineNotification } from '@nethserver/vue-tailwind-lib'
+import { getStandaloneRoutePrefix } from '@/lib/router'
+
+//// move to library
+
+const props = defineProps({
```
Tbaile, line 13: `props` appears not to be used, can it be deleted?

src/components/standalone/dashboard/RealTimeTrafficCard.vue
```diff
@@ -0,0 +1,94 @@
...
+
+function formatTraffic(value: number) {
+  value = Number(value.toFixed(2))
+  return byteFormat1000(value) + '\\s'
+}
+
+async function getTopTalkers() {
+  // show skeleton only the first time
+  if (!topTalkersIntervalId.value) {
+    loading.value.getTopTalkers = true
```
Tbaile, line 61: If this is only for the first time, you could just set the value to `true` on line 32

src/components/standalone/dashboard/RealTimeTrafficCard.vue
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  try {
+    const res = await ubusCall('ns.talkers', 'list', {
+      limit: NUM_HOSTS
+    })
+    topTalkers.value = res.data.talkers.map((row: any) => {
+      return { host: row.host, traffic: formatTraffic(row.totals.bandwidth) }
+    })
+  } catch (err: any) {
```
Tbaile, line 71: `err` is instance of `AxiosError`

src/components/standalone/dashboard/ServiceCard.vue
```diff
@@ -0,0 +1,171 @@
...
+
+  if (counterIntervalId.value) {
+    clearInterval(counterIntervalId.value)
+  }
+})
+
+async function getServiceStatus() {
+  // show skeleton only the first time
+  if (!statusIntervalId.value) {
+    loading.value.getServiceStatus = true
```
Tbaile, line 72: same as RealTimeTrafficChart, this could be set true by default

src/components/standalone/dashboard/SystemInfoCard.vue
```diff
@@ -0,0 +1,103 @@
...
+onUnmounted(() => {
+  if (systemInfoIntervalId.value) {
+    clearInterval(systemInfoIntervalId.value)
+  }
+})
+
+async function getSystemInfo() {
+  // show skeleton only the first time
+  if (!systemInfoIntervalId.value) {
+    loading.value.getSystemInfo = true
```
Tbaile, line 44: See above

src/components/standalone/dashboard/TrafficSummaryChart.vue
```diff
@@ -0,0 +1,88 @@
...
+  indexAxis: 'y',
+  scales: {
+    x: {
+      ticks: {
+        // format traffic on chart ticks
+        callback: function (value: number, index: number, ticks: any[]) {
+          return byteFormat1024(value)
+        },
+        // font color according to theme: gray-200 or gray-700
+        color: themeStore.isLight ? '#374151' : '#e5e7eb',
```
Tbaile, line 45: Instead of hard coding colors, we could reference tailwind configuration directly from javascript.
It's just a suggestion, never used it :shrug:  
Ref: https://tailwindcss.com/docs/configuration#referencing-in-java-script

Tbaile (commented):
Few things I would like to address:
- Some files are not run against prettier and eslint (minor chore, but useful one)
- There are some types set to `any` not for a meaningful reason, ( i.e. `ubuscall` always returns an `AxiosError` in case of exception) we just lose IDE support and syntactic sugar
-  Is `console.log` really necessary when a network error happens? It will still be printed in the console as a network error

Tbaile (approved):
linter goes: **BRRRR**

:smile:

src/components/standalone/dashboard/ServiceCard.vue
andre8244: In this case this doesn't work, since some cards need to retrieve just the status, while others need to retrieve just a counter, and they will never set `loading.getServiceStatus` or `loading.getServiceCounter' to `false`

src/components/standalone/dashboard/TrafficSummaryChart.vue
andre8244: It doesn't seem to work, anyway I'd stick with the actual implementation considering this:
image

andre8244:
> linter goes: **BRRRR**
> 
> 😄

After this commit linter is chilling :sunglasses: https://github.com/NethServer/nethsecurity-ui/pull/28/commits/275d50d554b1cca4f5a7813f7662c5624be730b6

PR #40, opened 2023-09-21
Tooltips support

andre8244:
- Install and configure [vue-tippy](https://github.com/KABBOUCHI/vue-tippy)
- Update `@nethserver/vue-tailwind-lib` to get tooltip support on various components

See https://github.com/NethServer/vue-tailwind-lib/pull/10

### DIFF
--- a/src/assets/main.css
+++ b/src/assets/main.css
@@ -62,3 +62,20 @@ code {
 hr {
   @apply border-gray-200 dark:border-gray-700;
 }
+
+/* tailwind theme for tooltip library */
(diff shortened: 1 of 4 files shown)


src/assets/main.css
Tbaile, line 1: So there's no CSS file in library? Shall we investigate that in NethServer/vue-tailwind-lib#10 before proceeding?

andre8244: We can improve this in the future, for the moment let's use a KISS approach

PR #41, opened 2023-09-25
Handle validation errors on ubusCall

andre8244:
Detect validation errors while invoking `/ubus/call` and throw a `NeValidationError` to properly handle the error validations on the UI.

See: https://github.com/NethServer/nethsecurity/pull/163

### DIFF
--- a/src/lib/standalone/ubus.ts
+++ b/src/lib/standalone/ubus.ts
@@ -4,20 +4,55 @@
 import axios from 'axios'
 import { getStandaloneApiEndpoint } from '../config'
 import { useLoginStore } from '@/stores/standalone/standaloneLogin'
+import { MessageBag } from '../validation'
+
+type ValidationError = {
+  message: string
+  parameter: string
+  value: string
+}
+
+export class NeValidationError extends Error {
+  errorBag: MessageBag
+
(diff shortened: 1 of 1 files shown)


src/lib/standalone/ubus.ts
```diff
@@ -4,20 +4,55 @@
...
 import { useLoginStore } from '@/stores/standalone/standaloneLogin'
+import { MessageBag } from '../validation'
+
+type ValidationError = {
+  message: string
+  parameter: string
+  value: string
+}
+
+export class NeValidationError extends Error {
```
Tbaile, line 15: I'd rather change this in:

```suggestion
type ValidationErrorResponse = {
  message: string
  parameter: string
  value: string
}

export class ValidationError extends Error {
```

just to keep things without explicit branding since we don't actually need it.
What do you think?

andre8244: I put the prefix thinking that we might move it to the library someday, but for now let's follow your suggestion

Tbaile: Don't think it will, many systems think differently on how to manage Validation Errors, better to keep business logic outside of a all-rounded library for components

PR #56, opened 2023-10-11
Port forward: ui fixes and better error handling

kevincela:
- add error handling for fetching firewall config inside CreateOrEditPortForwardDrawer
- move destination zone combobox inside advanced settings in CreateOrEditPortForwardDrawer
- replace empty cards in PortForward page with NeEmptyState
- clear errors when exiting CreateOrEditPortForwardDrawer

### DIFF
--- a/src/components/standalone/firewall/CreateOrEditPortForwardDrawer.vue
+++ b/src/components/standalone/firewall/CreateOrEditPortForwardDrawer.vue
@@ -165,6 +165,11 @@ watchEffect(() => {
 })
 
 function close() {
+  error.value.notificationTitle = ''
+  error.value.notificationDescription = ''
+  validationErrorBag.value.clear()
+  restrictIPValidationErrors.value = []
+
   resetForm()
   emit('close')
 }
--- a/src/components/standalone/firewall/CreateOrEditPortForwardDrawer.vue
+++ b/src/components/standalone/firewall/CreateOrEditPortForwardDrawer.vue
@@ -171,7 +171,9 @@ function close() {
 
 onMounted(() => {
   fetchOptions()
-  firewallConfig.fetch()
+  if (firewallConfig.loading || firewallConfig.error) {
+    firewallConfig.fetch()
+  }
 })
 
 function resetRestrictIPValidationErrors() {
@@ -286,13 +288,17 @@ async function createOrEditPortForward() {
     "
   >
     <NeInlineNotification
-      v-if="error.notificationTitle"
-      :title="error.notificationTitle"
-      :description="error.notificationDescription"
+      v-if="error.notificationTitle || firewallConfig.error"
+      :title="firewallConfig.error ? t('error.cannot_retrieve_zones') : error.notificationTitle"
+      :description="
+        firewallConfig.error
+          ? t(getAxiosErrorMessage(firewallConfig.error))
+          : error.notificationDescription
+      "
       class="mb-6"
       kind="error"
     />
-    <NeSkeleton v-if="loading || firewallConfig.loading" :lines="10" />
+    <NeSkeleton v-if="loading || firewallConfig.loading || firewallConfig.error" :lines="10" />
     <div v-else class="flex flex-col gap-y-6">
       <NeToggle :label="t('standalone.port_forward.status')" v-model="enabled" />
       <NeTextInput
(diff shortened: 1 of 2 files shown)


src/components/standalone/firewall/CreateOrEditPortForwardDrawer.vue
```diff
@@ -286,13 +298,17 @@ async function createOrEditPortForward() {
...
+      :description="
+        firewallConfig.error
+          ? t(getAxiosErrorMessage(firewallConfig.error))
+          : error.notificationDescription
+      "
       class="mb-6"
       kind="error"
     />
-    <NeSkeleton v-if="loading || firewallConfig.loading" :lines="10" />
+    <NeSkeleton v-if="loading || firewallConfig.loading || firewallConfig.error" :lines="10" />
```
andre8244, line 311: We usually don't show a skeleton if an error is displayed

src/components/standalone/firewall/CreateOrEditPortForwardDrawer.vue
```diff
@@ -165,13 +165,25 @@ watchEffect(() => {
...
+  restrictIPValidationErrors.value = []
+
   resetForm()
   emit('close')
 }
 
 onMounted(() => {
   fetchOptions()
-  firewallConfig.fetch()
+  if (firewallConfig.loading || firewallConfig.error) {
```
andre8244, line 184: Can you explain this condition?

kevincela: Considering how `useFirewallStore` is a global store and is used in different pages other than the port forward one, I thought it wouldn't make much sense to refetch its items when they already loaded, hence why i've included this condition. If you think this may cause issues I can remove it though!

src/components/standalone/firewall/CreateOrEditPortForwardDrawer.vue
kevincela: Hmm, this is true but doing it this way could also cause a scenario where the user can interact with the form even if the firewallConfig fetch did not go as planned... How should we handle this case?

src/components/standalone/firewall/CreateOrEditPortForwardDrawer.vue
```diff
@@ -165,13 +165,25 @@ watchEffect(() => {
 })
 
 function close() {
+  if (
```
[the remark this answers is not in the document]
kevincela: Not strictly needed, but if the error is given by a failed request of the form's options then clearing it could be confusing if the user reopens it again (they would see no error notification, but the options still aren't available). Maybe at this point we should simply refetch the options every time the drawer is opened, in case of an error?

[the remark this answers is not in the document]
andre8244: Yes, if the drawer need some data we usually fetch it every time it opens, so we can get updated data. If the code doesn't currently work this way we can postpone these changes to another PR, it's not an essential feature at the moment.

src/components/standalone/firewall/CreateOrEditPortForwardDrawer.vue
andre8244: > Hmm, this is true but doing it this way could also cause a scenario where the user can interact with the form even if the firewallConfig fetch did not go as planned... How should we handle this case?

If an error occurs while retrieving drawer data we usually display only the error and hide the form

src/components/standalone/firewall/CreateOrEditPortForwardDrawer.vue
andre8244: > Considering how `useFirewallStore` is a global store and is used in different pages other than the port forward one, I thought it wouldn't make much sense to refetch its items when they already loaded, hence why i've included this condition. If you think this may cause issues I can remove it though!

Now it's clear, thanks

src/components/standalone/firewall/CreateOrEditPortForwardDrawer.vue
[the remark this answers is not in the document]
kevincela: Ok, I've updated the drawer so that it refetches the data every time it is shown, now it should be ok. I've also removed the aforementioned condition because of this change.

src/components/standalone/firewall/CreateOrEditPortForwardDrawer.vue
kevincela: Ok, now if there's an error while fetching the firewall config the form does not show. I've also removed the condition from the skeleton.

andre8244 (approved):

PR #58, opened 2023-10-11
DPI filter

andre8244:
- DPI filter page
- Fix `ServiceCard` component
- Minor fixes

Card: https://trello.com/c/iYYmX9mZ/136-dpi-ui


src/components/standalone/dpi/DpiRules.vue
Tbaile, line 1: Every `reloadData` does not in fact reload the changes to UCI.

Tbaile (changes requested):

Tbaile (approved):

PR #72, opened 2023-10-23
Add OpenVPN tunnel page

kevincela:
- Add OpenVPN tunnel page and related translations and icons

Depends on NethServer/nethsecurity#203


src/components/standalone/openvpn_tunnel/TunnelTable.vue
```diff
@@ -0,0 +1,218 @@
...
+}>()
+
+const emit = defineEmits([
+  'tunnel-delete',
+  'tunnel-edit',
+  'tunnel-toggle-enable',
+  'tunnel-download'
+])
+
+const tableHeaders = [
```
Tbaile, line 21: I feel like this is a bit complex, it's possible to add a some comments to explain why and what does this do? I'm feeling we're gonna lose why there's this procedural generation of the headers fields

src/components/standalone/openvpn_tunnel/TunnelTable.vue
Tbaile, line 1: Is there any close term usage of the `connection` commented column? If not, can we remove it?

src/components/standalone/openvpn_tunnel/DownloadTunnelModal.vue
```diff
@@ -0,0 +1,118 @@
...
+  'nethsecurity_client_configuration' | 'private_key_tunnel_ca_certificates' | ''
+>('')
+
+const downloadOptions = [
+  {
+    id: 'nethsecurity_client_configuration',
+    label: t('standalone.openvpn_tunnel.nethsecurity_client_configuration'),
+    description: t('standalone.openvpn_tunnel.nethsecurity_client_configuration_description')
+  }
+  /*{
```
Tbaile, line 34: This can be safely removed, there's the TODO already present down below

public/i18n/en/translation.json
```diff
@@ -814,6 +830,81 @@
     "vpn": {
       "title": "VPN"
     },
+    "openvpn_tunnel": {
+      "title": "OpenVPN tunnel",
+      "tabs": {
+        "server_tunnel": "Server tunnel",
+        "client_tunnel": "Client tunnel"
+      },
+      "server_tunnel_description": "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Suspendisse gravida, nibh vel finibus suscipit, mauris erat porta",
```
gsanchietti, line 839:
```suggestion
      "server_tunnel_description": "On this page, you can configure the server for OpenVPN tunnels. OpenVPN tunnels are used to securely connect remote networks. After the configuration is complete, you can export the settings, which can be imported into another NethSecurity device from the OpenVPN client section.",
```

public/i18n/en/translation.json
```diff
@@ -814,6 +830,81 @@
...
+      "client_tunnel_description": "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Suspendisse gravida, nibh vel finibus suscipit, mauris erat porta",
```
gsanchietti, line 840:
```suggestion
      "client_tunnel_description": "On this page, you can import a configuration file generated from another NethSecurity's OpenVPN tunnel server page. You can also manually configure a client to connect it to any device that supports OpenVPN.",
```

src/components/standalone/openvpn_tunnel/CreateOrEditTunnelDrawer.vue
```diff
@@ -0,0 +1,846 @@
...
+        label: t('standalone.openvpn_tunnel.auto'),
+        description: `(${t('standalone.openvpn_tunnel.server_client_negotiation')})`
+      },
+      ...(await ubusCall('ns.ovpntunnel', 'list-digest')).data.digests.map(
+        (digest: { name: string; description: string }) => ({
+          id: digest.name,
+          label: digest.name
+        })
+      )
+    ]
```
Tbaile, line 234: Good as one-liners, not so good at readability, I think a separation of these or additional comments are needed

src/components/standalone/openvpn_tunnel/CreateOrEditTunnelDrawer.vue
```diff
@@ -0,0 +1,846 @@
...
+      ? ((
+          await ubusCall('ns.ovpntunnel', 'get-tunnel-client', {
+            id: props.itemToEdit.id
+          })
+        ).data as ClientTunnelPayload)
+      : ((
+          await ubusCall('ns.ovpntunnel', 'get-tunnel-server', {
+            id: props.itemToEdit.id
+          })
+        ).data as ServerTunnelPayload)
```
Tbaile, line 262: I think an `if` approach is needed here

public/i18n/en/translation.json
```diff
@@ -814,6 +830,81 @@
...
+      "authentication": "Authentication",
+      "username": "Username",
+      "password": "Password",
+      "certificate": "Certificate",
+      "username_password_certificate": "Username, password and certificate",
+      "mode": "Mode",
+      "bridged": "Bridged",
+      "routed": "Routed",
+      "could_not_parse_configuration_file": "Could not parse configuration file",
+      "public_endpoints_tooltip": "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Suspendisse gravida, nibh vel finibus suscipit, mauris erat porta",
```
gsanchietti, line 902:
```suggestion
      "public_endpoints_tooltip": "Public endpoints are a list of IP addresses or hostnames that clients can use to reach the OpenVPN tunnel server.",
```

public/i18n/en/translation.json
```diff
@@ -814,6 +830,81 @@
...
+      "local_networks_tooltip": "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Suspendisse gravida, nibh vel finibus suscipit, mauris erat porta",
```
gsanchietti, line 903:
```suggestion
      "local_networks_tooltip": "Local networks are a list of network addresses that need to be reachable from the other end of the OpenVPN tunnel.",
```

Tbaile (changes requested):

public/i18n/en/translation.json
```diff
@@ -814,6 +830,81 @@
...
+      "remote_networks_tooltip": "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Suspendisse gravida, nibh vel finibus suscipit, mauris erat porta",
```
gsanchietti, line 904:
```suggestion
      "remote_networks_tooltip": "Remote networks are a list of network addresses present at the other end of the OpenVPN tunnel. The networks specified here will be added as static routes in the firewall when the VPN is active.",
```

public/i18n/en/translation.json
```diff
@@ -814,6 +830,81 @@
...
+      "extra_remote_networks_tooltip": "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Suspendisse gravida, nibh vel finibus suscipit, mauris erat porta",
```
gsanchietti, line 905:
```suggestion
      "extra_remote_networks_tooltip": "When the VPN is active, these network addresses will be added as static routes regardless of the configuration received from the OpenVPN server.",
```

public/i18n/en/translation.json
```diff
@@ -814,6 +830,81 @@
...
+      "topology_tooltip": "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Suspendisse gravida, nibh vel finibus suscipit, mauris erat porta"
```
gsanchietti, line 906: Is it possible to format this with a little bit of html code?
```suggestion
      "topology_tooltip": "Subnet (recommended): This method offers a straightforward configuration approach. Instead of using a point-to-point setup, utilize a subnet configuration by specifying a local IP address and subnet mask for the tun interface. In this mode, each connected client is assigned a unique IP address, and it is compatible with Windows systems as well. P2P: Implement a point-to-point topology where the client's tun interface remote endpoint consistently points to the server's tun interface local endpoint. This configuration assigns a single IP address to each connected client. Utilize this mode exclusively when none of the connected clients operate on Windows systems."
```

kevincela: I've separated the tooltip translation into distinct elements to allow styling without formatting raw HTML code from the translations, should be ok now!

src/components/standalone/openvpn_tunnel/CreateOrEditTunnelDrawer.vue
kevincela: Tried adding some comments to better describe the lists' components. Let me know if this can work (we could also simply separate the components, but in this case I think a comment should be enough)

src/components/standalone/openvpn_tunnel/TunnelTable.vue
kevincela: Added some comments to better describe the optional inclusion of headers based on the tunnel type

Tbaile (approved):

src/components/standalone/openvpn_tunnel/CreateOrEditTunnelDrawer.vue
Tbaile: It should be fine, I was suggesting it just so that we don't have to ping pong back and fourth from the back-end doc to the UI

PR #74, opened 2023-10-26
Fix custom zones

andre8244:
- Fix interfaces configuration on custom zones
- Allow editing of special zones (lan, wan...)
- Fix icon and colors of custom zones

Other changes:
- Fix i18n string


src/lib/fontawesome.ts
Tbaile, line 1: I smell a version mismatch underneath the `@ts-ignore` workaround.
All icons are `IconDefinition` like every other, however only the newly imported one won't work, this smells fishy and can cause issues on the long run.

Tbaile (changes requested):

src/lib/fontawesome.ts
Tbaile: Moved issue to #75

Tbaile (approved):

PR #75, opened 2023-10-30
feat(dpi): add some icons for popular apps

Tbaile:


src/lib/fontawesome.ts
Tbaile, line 1: I smell a version mismatch underneath the @ts-ignore workaround.
All icons are IconDefinition like every other, however only the newly imported one won't work, this smells fishy and can cause issues on the long run.

Tbaile:
Superseded by https://github.com/NethServer/nethsecurity-ui/pull/89

PR #86, opened 2023-11-13
Add update page

kevincela:
- Add update page
- Add `useTimer` composable to implement timers (used in the update dialogs)

Depends on https://github.com/NethServer/nethsecurity/pull/220

Card: https://trello.com/c/kFmdJbTU/182-updates-page


src/composables/useTimer.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export type TimerOptions = {
+  duration: number
+  progressStep: number
+  onProgressStep?: (currentProgress: number) => void
+  onTimerFinish?: () => void
+}
+
+/**
+ * Composable that enables the creation of a timer, and the end of which an action is executed.
```
andre8244, line 11:
```suggestion
 * Composable that enables the creation of a timer, at the end of which an action is executed.
```

src/lib/fontawesome.ts
```diff
@@ -200,4 +201,5 @@ export async function loadFontAwesome(app: any) {
   library.add(fasTrafficCone)
   library.add(fasStar)
   library.add(faKey)
+  library.add(faCalendar)
```
andre8244, line 204: We can remove this icon

src/views/standalone/system/UpdateView.vue
```diff
@@ -0,0 +1,325 @@
...
+    v-if="error.notificationTitle"
+    :title="error.notificationTitle"
+    :description="error.notificationDescription"
+    class="my-4"
+    kind="error"
+    ><template #details v-if="error.notificationDetails">
+      {{ error.notificationDetails }}
+    </template></NeInlineNotification
+  >
+  <NeSkeleton v-if="loading" :lines="10" />
```
andre8244, line 146: Minor skeleton improvement: can we show the title and descriptions of the two sections ("Bug & security fixes" and "System update") while loading data? The skeleton can appear only at the right column of the page

andre8244 (commented):
Just some minor improvements

PR #90, opened 2023-11-16
feat: show toast notification on api error

andre8244:
Show a toast notification whenever an API call fails. By interacting with the toast notification it is be possible to:
- copy the cURL command equivalent to the failed API invocation
- show a modal that displays info about the API call:
  - error message
  - HTTP method and URL
  - request payload (if available)
  - response payload (if available)

Other changes: add `@types/lodash-es` to dependencies

Related PR: https://github.com/NethServer/vue-tailwind-lib/pull/21


src/stores/standalone/notifications.ts
```diff
@@ -0,0 +1,128 @@
...
+      : DEFAULT_NOTIFICATION_TIMEOUT
+  }
+
+  const createNotificationFromAxiosError = (axiosError: any) => {
+    const notificationTitle = axiosError.config.url.includes('/ubus/call')
+      ? t('notifications.ubus_call_failed')
+      : t('notifications.api_call_failed')
+
+    const notification: NeNotification = {
+      id: uid(),
```
Tbaile, line 43: May I suggest instead of using UIDs (and an additional library), just using an incremental ID?
Example: https://github.com/NethServer/ns8-user-manager/blob/main/src/stores/useNotificationEngine.ts

andre8244: I'd prefer to stick with random IDs, notification will be soon maintained inside local storage so we cannot reset the incremental ID anytime the page is refreshed

Tbaile (approved):

PR #97, opened 2023-11-23
fix: update system with image file by using drag&drop

andre8244:
Use an updated version of `NeFileInput` component inside **System / Update** to fix image file selection by drag & drop.

Other changes: rename function `uploadImageAndReboot` to `updateImageAndReboot`

Related PR: https://github.com/NethServer/vue-tailwind-lib/pull/22


src/components/standalone/update/UploadImageDrawer.vue
```diff
@@ -32,17 +32,10 @@ const uploadedImageFileName = ref('')
 
 const fileToUpload = ref<File | null>(null)
 
-async function handleFileUpload(event: Event) {
+async function handleFileUpload() {
   fileInputValidationError.value = ''
   isReadyToUpdate.value = false
   isUploadingImage.value = false
```
kevincela, line 38:
```suggestion
  isUploadingImage.value = false
  if (!fileToUpload.value) return
```
A check on the uploaded file should probably stll be kept, since there are cases where the file can be empty (for example, when the user cancels the file selection window). I don't think it makes much sense validating the file when it's not even selected

src/components/standalone/update/UploadImageDrawer.vue
```diff
@@ -119,7 +112,7 @@ async function uploadImageAndReboot() {
       <div>
         <NeFileInput
           :label="`${t('standalone.update.image_file')} (*.img.gz)`"
-          @change="handleFileUpload"
+          @select="handleFileUpload"
           :invalid-message="fileInputValidationError"
```
kevincela, line 116: I've just noticed this component doesn't have the `dropzoneLabel`, which should be required. We should probably add it (and now that i think about it, maybe it would be better to use a single shared `upload_file_description` translation entry, since we already have two identical dropzone labels for the other `NeFileInput`s used in the ui

kevincela (commented):
As a sidenote: the `select` emit of the `NeFileInput` for some reason does not seem to be working on my end, even after updating the package. However, when using a copy of the component inside of the project, it seems to be working fine. I wonder if there is some issue in the published package?

kevincela (approved):

Tbaile (approved):

PR #103, opened 2023-11-29
fix: temporary workaround for overflow issue

Tbaile:
Even if this allows the user to eventually scroll for an option in a dropdown it's still not 100% resolutive (ui looks bad, really bad).

Only soution found is to remove entirely the outer box (we will lose rounded corners, so it needs a good revamp)

Table will be removed from here and eventually added to the component library, in that time it's a good approach to refactor the table styling to fix the `overflow-y-even-if-not-damn-set` once and for all.

### DIFF
--- a/src/components/standalone/NeTable.vue
+++ b/src/components/standalone/NeTable.vue
@@ -145,7 +145,7 @@ table {
 }
 
 .table-basic thead {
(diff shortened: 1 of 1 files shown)


Tbaile:
Replaced by: NethServer/vue-tailwind-lib#26

PR #104, opened 2023-11-29
refactor: move interfaces and devices logic to backend

andre8244:
Refactor **Interfaces and devices** page: move logic to the new API `ns.devices`

See: https://github.com/NethServer/nethsecurity/pull/239


src/stores/standalone/firewall.ts
```diff
@@ -146,6 +162,28 @@ export class Forwarding {
...
+    rank2 = 99
+  }
+
+  if (rank1 < rank2) {
+    return -1
+  }
+  if (rank1 > rank2) {
+    return 1
+  }
+  return 0
```
Tbaile, line 184: Isn't here better to return `rank1 - rank2`?

src/components/standalone/NeAxiosErrorModal.vue
```diff
@@ -4,6 +4,8 @@ import { computed, ref, watch } from 'vue'
 import { useI18n } from 'vue-i18n'
 import { useNotificationsStore } from '@/stores/standalone/notifications'
 
+//// move to lib after removing ubus references
```
Tbaile, line 7: I wouldn't move this component in the library, it's too much tailored to the specific use than anything else

Tbaile (approved):

PR #118, opened 2023-12-19
feat(devices): read allowed zone from backend

andre8244:
Read the list of zones that can be selected for device configuration from the backend, using `ns.devices list-zones-for-device-config` API.

See also: https://github.com/NethServer/nethsecurity/pull/254


src/components/standalone/interfaces_and_devices/ConfigureDeviceDrawer.vue
```diff
@@ -115,6 +113,7 @@ let bondingPolicy = ref('balance-rr')
 let bondingPolicyRef = ref<HTMLDivElement | null>()
 let bondPrimaryDevice = ref('')
 let bondPrimaryDeviceRef = ref<HTMLDivElement | null>()
+let allowedZones = ref<any>([])
```
Tbaile, line 116:
```suggestion
let allowedZones = ref<Array<any>>([])
```

andre8244: :innocent:

Tbaile (approved):

PR #129, opened 2023-12-22
chore: import NeCombobox and NeBadge from vue-components

andre8244:
- Import NeCombobox and NeBadge components from `vue-components` instead of `vue-tailwind-lib`
- Pass i18n props to NeCombobox. Before this change, English strings where used as a fallback whenever i18n props were not passed to the component.


andre8244:
The build will succeed after importing the new version of `vue-components`

Tbaile (approved):

PR #144, opened 2024-01-16
Devices and interfaces: move VLAN device type under advanced settings

kevincela:
- Move VLAN device type under advanced settings in the create VLAN device drawer

Card: https://trello.com/c/nNY0G6HV/221-devices-and-interfaces-improvement-on-vlan

### DIFF
--- a/public/i18n/en/translation.json
+++ b/public/i18n/en/translation.json
@@ -660,7 +660,8 @@
       "no_devices_available": "No devices available",
       "custom_zone": "Custom",
       "cannot_remove_device_configuration": "Cannot remove device configuration",
-      "hotspot_network": "Network"
+      "hotspot_network": "Network",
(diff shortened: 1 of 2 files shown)


public/i18n/en/translation.json
```diff
@@ -660,7 +660,8 @@
       "no_devices_available": "No devices available",
       "custom_zone": "Custom",
       "cannot_remove_device_configuration": "Cannot remove device configuration",
-      "hotspot_network": "Network"
+      "hotspot_network": "Network",
+      "advanced_settings": "Advanced settings"
```
andre8244, line 664: We can use `common.advanced_settings`

kevincela: Whoops I forgot about it, should be fixed now

PR #145, opened 2024-01-16
DNS: add specific validation for DNS forwarding servers

kevincela:
- Add specific validation function for DNS forwarding servers, following the same specs used by LuCI

Card: https://trello.com/c/dyu6tLPV/273-dns-relax-validator-for-dns-forwarding-servers

### DIFF
--- a/public/i18n/en/translation.json
+++ b/public/i18n/en/translation.json
@@ -117,6 +117,7 @@
     "password_number_required": "Password must contain at least one number",
     "password_special_character_required": "Password must contain at least one special character",
     "invalid_equal": "Passwords don't match",
+    "invalid_dns_forwarding": "Enter a valid IP address or DNS forwarding item",
     "user_not_found": "User not found",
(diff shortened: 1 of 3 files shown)


kevincela:
@gsanchietti which change would you propose on the tooltip? The current value is as follows:

```
      "dns_forwarding_servers_tooltip": "Configure upstream DNS server IP addresses to route non-local domain queries to these servers.",
```

gsanchietti:
> @gsanchietti which change would you propose on the tooltip? The current value is as follows:
> 
> ```
>       "dns_forwarding_servers_tooltip": "Configure upstream DNS server IP addresses to route non-local domain queries to these servers.",
> ```

I'd go with this:
```
Configure upstream DNS server IP addresses to route non-local domain queries to these servers. Enter the IP address of the remote DNS server. Use '/<domain>/<server>' syntax to route only requests for a specific domain to the given server. Use the '/<server>#<port>` syntax if the upstream server does not listen on a standard port.
```

Tbaile (approved):

PR #150, opened 2024-01-19
refactor: bond management processes

Tbaile:
UI is quite intricate, but here's what I've tried to accomplish so far:
- deleting the `zone` param when creating a bond, allows to configure the bond according to user preference, also removed the radio selection entirely for bonds
- bond naming is **weird** (and it's a compliment). When you define a named bond, the software first creates a `bonding-<name>` device for said bond, then when device gets up it's renamed to `bond-<name>`. IF THE `bond-<name>` EXCEEDS THE 15 CHARACTERS IT WILL BE STUCK IN THE FIRST STATE. Hence there's a custom check for bond name length.
- Added delete bond modal using custom call from API
- Due to changes in the API responses, there's an additional `bond_interface` field in the DeviceOrIface interface.
- Added additional check in `isBond` function, checking for name too. (can be change to check if there's a `bond_interface` field tho)
- Filtering out from the response the bond interfaces if there's a bond device, might be overkill now since the API already filters them.
- If you configure a bond, trick the drawer to make it believe that it's configuring a physical device with the correct device name. **NOTE**: It seems it's bugged in some way, need assistance on where it might be different from other configuration drawers.
- slicing name in rows and removing `bond-` prefix.
- few revisiting on the buttons for the rows, this might have introduced bugs, adding configure/edit/delete of bond.

Prerequisites:
- https://github.com/NethServer/nethsecurity/pull/277

### DIFF
--- a/src/components/standalone/interfaces_and_devices/ConfigureDeviceDrawer.vue
+++ b/src/components/standalone/interfaces_and_devices/ConfigureDeviceDrawer.vue
@@ -325,6 +325,20 @@ const drawerTitle = computed(() => {
 })
 
 watch(
+  logicalIfaceType,
+  () => {
+    if (logicalIfaceType.value == 'bond') {
+      const firstRandom = Math.floor(Math.random() * 255)
+      const secondRandom = Math.floor(Math.random() * 255)
+      ipv4Address.value = `127.${firstRandom}.${secondRandom}.1/30`
+    } else {
+      ipv4Address.value = ''
+    }
+  },
+  { immediate: true }
+)
+
+watch(
   () => props.isShown,
   () => {
     if (props.isShown) {
--- a/src/components/standalone/interfaces_and_devices/ConfigureDeviceDrawer.vue
+++ b/src/components/standalone/interfaces_and_devices/ConfigureDeviceDrawer.vue
@@ -479,6 +479,7 @@ function prepareConfigureDeviceData() {
   if (props.deviceType === 'logical' && logicalIfaceType.value === 'bond') {
     data.bonding_policy = bondingPolicy.value
     data.bond_primary_device = bondPrimaryDevice.value
+    delete data.zone
   }
 
   if (protocol.value === 'pppoe') {
@@ -586,7 +587,11 @@ function validate() {
     } else {
       // check syntax
       {
-        let { valid, errMessage, i18Params } = validateUciName(interfaceName.value, 15)
+        let maxLen = 15
+        if (logicalIfaceType.value == 'bond') {
+          maxLen = 10
+        }
+        let { valid, errMessage, i18Params } = validateUciName(interfaceName.value, maxLen)
         if (!valid) {
           error.value.interfaceName = t(errMessage as string, i18Params as any)
           if (isValidationOk) {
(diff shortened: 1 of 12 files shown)


src/components/standalone/interfaces_and_devices/ConfigureDeviceDrawer.vue
```diff
@@ -288,6 +293,51 @@ const devicesUsedByOhterBridgesOrBonds = computed(() => {
...
+  }
+})
+
+watch(
+  logicalIfaceType,
+  () => {
+    if (logicalIfaceType.value == 'bond') {
+      const firstRandom = Math.floor(Math.random() * 255)
+      const secondRandom = Math.floor(Math.random() * 255)
+      ipv4Address.value = `127.${firstRandom}.${secondRandom}.1/30`
```
gsanchietti, line 333:
```suggestion
      ipv4Address.value = `127.${firstRandom}.${secondRandom}.1/32`
```

src/components/standalone/interfaces_and_devices/ConfigureDeviceDrawer.vue
```diff
@@ -288,6 +293,51 @@ const devicesUsedByOhterBridgesOrBonds = computed(() => {
...
+    return t('standalone.interfaces_and_devices.create_logical_interface')
+  }
+})
+
+watch(
+  logicalIfaceType,
+  () => {
+    if (logicalIfaceType.value == 'bond') {
+      const firstRandom = Math.floor(Math.random() * 255)
+      const secondRandom = Math.floor(Math.random() * 255)
```
gsanchietti, line 332: Avoid to choose 0.

src/components/standalone/interfaces_and_devices/ConfigureDeviceDrawer.vue
```diff
@@ -922,12 +990,24 @@ async function listZonesForDeviceConfig() {
...
           <NeTextInput
-            :label="t('standalone.interfaces_and_devices.ipv4_address_cidr')"
+            :label="
+              logicalIfaceType === 'bond'
+                ? t('standalone.interfaces_and_devices.management_ipv4_address_cidr')
+                : t('standalone.interfaces_and_devices.ipv4_address_cidr')
+            "
             v-model.trim="ipv4Address"
             :invalidMessage="t(error.ipv4Address)"
             :disabled="loading.configure"
```
gsanchietti, line 1000: Why not disabling it to avoid errors from the user?
```suggestion
          :disabled="loading.configure || logicalIfaceType === 'bond'"
```

gsanchietti (changes requested):

src/components/standalone/interfaces_and_devices/ConfigureDeviceDrawer.vue
Tbaile: Haven't thought of that. Nice.

src/components/standalone/interfaces_and_devices/ConfigureDeviceDrawer.vue
Tbaile: [Following official documentation for .random() usage](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/random#getting_a_random_integer_between_two_values), updating it now

gsanchietti (approved):

PR #151, opened 2024-01-19
Devices and interfaces: rework alias interface drawer to use NeMultiTextInput

kevincela:
- Use NeMultiTextInput on the create/edit alias interface drawer to maintain uniformity
- Simplify validation with the new inputs in mind
- Add option, in NeMultiTextInput, to focus on the first text-input if available

Card: https://trello.com/c/9OWsVjah/222-devices-and-interfaces-alias-improvement-ui

### DIFF
--- a/src/components/standalone/NeMultiTextInput.vue
+++ b/src/components/standalone/NeMultiTextInput.vue
@@ -103,7 +103,6 @@ function updateModelKey(idx: number, newValue: string) {
 }
 
 function focus() {
-  console.log(inputRef.value)
   inputRef.value?.focus()
 }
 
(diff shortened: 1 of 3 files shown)


src/components/standalone/NeMultiTextInput.vue
```diff
@@ -96,6 +102,11 @@ function updateModelKey(idx: number, newValue: string) {
   emitUpdate()
 }
 
+function focus() {
```
andre8244, line 105: This is fine, for the next time we can use [focusElement function from the UI library](https://github.com/NethServer/vue-tailwind-lib/blob/3f4bfe8f605bb393ea950be80b71dc81afeb0d7c/src/lib/utils.ts#L26)

src/components/standalone/NeMultiTextInput.vue
```diff
@@ -96,6 +102,11 @@ function updateModelKey(idx: number, newValue: string) {
...
+  console.log(inputRef.value)
```
andre8244, line 106: Let's remove this debug print

andre8244 (approved):

PR #153, opened 2024-01-24
Two-factor authentication

andre8244:
- Add 2FA management inside **Account settings** page
- Request OTP code in loging page if 2FA is enabled

See also: https://github.com/NethServer/nethsecurity-api/pull/3


src/components/standalone/FormLayout.vue
Tbaile, line 1: Why such change?

src/components/standalone/StandaloneAppLogin.vue
```diff
@@ -120,84 +151,183 @@ function validate() {
...
+  for (const [key, value] of Object.entries(error.value) as [string, any][]) {
+    if (typeof value === 'string') {
+      // @ts-ignore
+      error.value[key] = ''
+    } else if (Array.isArray(value)) {
+      // @ts-ignore
+      error.value[key] = []
+    }
+  }
+}
```
Tbaile, line 204: Definitely not agreeing on error handling this way, it's error prone at best on the long run

Consider separating validation errors (the ones that end up below input fields) using the messagebag
And adding an Error object to wrap any unforeseeable error and display them differently

src/lib/validation.ts
```diff
@@ -685,6 +708,23 @@ export class MessageBag extends Map<string, Array<string>> {
...
+
+  if (validationErrors) {
+    validationErrors.forEach((validationError: any) => {
+      const errorMessages = errorBag.get(validationError.parameter) || []
+      errorMessages.push(validationError.message)
+      errorBag.set(validationError.parameter, errorMessages)
+    })
+  }
+  return errorBag
+}
```
Tbaile, line 726: Shouldn't you check for a ValidationError instance? 
Please don't tell me the format is different for 2FA requests :pray:

Tbaile (changes requested):

src/components/standalone/FormLayout.vue
andre8244: It's a minor enhancement to preserve style while using slots

src/components/standalone/StandaloneAppLogin.vue
andre8244: This is just a quick way to empty `error` object, we do it like this in many other components. Refactoring error management for all components is out of the scope of this PR, but I'm gonna remove `clearErrors` function here

src/lib/validation.ts
andre8244: No. This function is the externalization of [these lines](https://github.com/NethServer/nethsecurity-ui/pull/153/files/1f6fd80c23ec3956677226cbfee2b0130a9350df#diff-2e3fe53b79c06a0f2eada22f9bc6ea7eabdbffaa67bc8760a94d930b68959586L40-L49). This way validation errors are now detected and handled also from API calls, not only ubus

src/components/standalone/StandaloneAppLogin.vue
Tbaile: True that this is not something that we do everywhere, but something that even remotely requires @ts-ignore is a real bummer, we lose even what the error we manage even is

Tbaile (approved):

PR #163, opened 2024-01-31
Migration: apply various UI enhancements

kevincela:
- Update migration in progress translation to inform the users of the IPs through which the machine will be reachable
- Rework migration in progress timer using the `useTimer` composable
- Minor migration drawer ui fixes (reset the form when the drawer is reoponed, avoid closing migration in progress modal)
- Avoid showing axios error toast when the migration call fails, even if successful

Card: https://trello.com/c/T27Pn7IP/294-improve-migration-ui


gsanchietti:
I've tested it and works quite well! I'll leave the review to @andre8244 or @Tbaile

Tbaile (approved):

PR #167, opened 2024-02-02
Port forward: allow ip cidr in restrict field, update translations

kevincela:
- Allow usage of IP addresses with CIDR notation in the restrict field when adding or editing port forwards
- Update restrict field labels ("Restrict access from" instead of "Restrict access to")


public/i18n/en/translation.json
```diff
@@ -53,9 +53,10 @@
     "invalid_ip_address": "Invalid IP address",
     "invalid_ip_v4_address": "Invalid IPv4 address",
     "invalid_ip_v6_address": "Invalid IPv6 address",
+    "invalid_ip_cidr_address": "Enter a valid IPv4 or IPv6 address, with optional CIDR notation",
     "invalid_cidr_v4_address": "Enter a valid IPv4 address with CIDR notation",
     "invalid_cidr_v6_address": "Enter a valid IPv6 address with CIDR notation",
-    "invalid_ip_v6_address_or_cidr_v6": "Enter a valid IPv6 address or IPv6 address with CIDR notation",
+    "invalid_ip_v6_address_or_cidr_v6": "Enter a valid IPv4 address or IPv6 address with CIDR notation",
```
andre8244, line 59: I don't understand this change, anyway this i18n string seems unused, so I guess we can remove it

public/i18n/en/translation.json
```diff
@@ -53,9 +53,10 @@
     "invalid_ip_address": "Invalid IP address",
     "invalid_ip_v4_address": "Invalid IPv4 address",
     "invalid_ip_v6_address": "Invalid IPv6 address",
+    "invalid_ip_cidr_address": "Enter a valid IPv4 or IPv6 address, with optional CIDR notation",
```
andre8244, line 56:
```suggestion
    "invalid_ip_address_or_cidr": "Enter a valid IPv4 or IPv6 address, with optional CIDR notation",
```

public/i18n/en/translation.json
kevincela: My bad, I didn't read the string properly! Yeah I guess we can simply delete the item

andre8244 (approved):

PR #182, opened 2024-02-19
fix: loading of Poppins font

andre8244:
In some network/DNS scenarios the UI might not be able to retrieve Poppins font from `fonts.googleapis.com`, showing a blank page for about ten seconds before loading the UI.

Two different solutions have been explored:
- Loading the font from `fonts.googleapis.com` asynchronously: `<link rel="preload" as="style" href="https://fonts.googleapis.com/css?family=Poppins&display=swap" rel="stylesheet" />` (note `rel="preload"`)
- Loading the font locally from a font file

The second solution has been chosen and implemented in this PR


Tbaile (approved):
Font is rather small, it's possible in the future to add it through [fontsource](https://fontsource.org/fonts/poppins), good for me

PR #186, opened 2024-02-20
Interfaces: display packets in a more readable format

kevincela:
- Display packets in a more readable format in the Devices and interfaces page: 
  - Value is formatted depending on the browser's locale
  - Use M unit if packets are above 1 million

Card: https://trello.com/c/dNrGfN8n/324-interfaces-humanify-packet-count

### DIFF
--- a/src/views/standalone/network/InterfacesAndDevicesView.vue
+++ b/src/views/standalone/network/InterfacesAndDevicesView.vue
@@ -790,10 +790,7 @@ function formatPackets(packets: number) {
                           <span class="font-medium">TX: </span>
                           <span>{{ getTxBytes(device) || '-' }}</span>
                           <span v-if="device.stats?.tx_packets">
-                            ({{
-                              device.stats.tx_packets ? formatPackets(device.stats.tx_packets) : '-'
-                            }}
-                            pkts)</span
+                            ({{ formatPackets(device.stats.tx_packets) }} pkts)</span
                           >
                         </div>
                       </div>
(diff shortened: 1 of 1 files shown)


src/views/standalone/network/InterfacesAndDevicesView.vue
```diff
@@ -775,14 +783,17 @@ function isDeviceConfigurable(deviceOrIface: DeviceOrIface) {
                           <span class="font-medium">RX: </span>
                           <span>{{ getRxBytes(device) || '-' }}</span>
                           <span v-if="device.stats?.rx_packets">
-                            ({{ device.stats.rx_packets }} pkts)</span
+                            ({{ formatPackets(device.stats.rx_packets) }} pkts)</span
```
andre8244, line 786: Is a null check on `device.stats.rx_packets` needed? This check is present for `tx_packets`

kevincela: That's a good question, initially I simply considered the original state of the view (which used the bool check only for `tx_packets`), but looking at the payload i'm not sure it's even needed, since I'm seeing it's always a number... I think we need to decide whether we simply want to show "0 pkts" for both fields or "- pkts" if no packets are sent/received. I'd personally use the first option

andre8244: Looking more carefully, the null check inside `{{ ... }}` is useless: the two `span`s won't appear because they contain a null check themselves: `<span v-if="device.stats?.rx_packets">` and `<span v-if="device.stats?.tx_packets">`. We can safely remove the null check on `tx_packets` at line 794

andre8244 (approved):

PR #204, opened 2024-03-05
chore: rename guests zone to guest

gsanchietti:
The correct name is guest, in singular form.


kevincela (commented):
Overall it seems to be okay, hovever I see that in `firewall.ts` line 184 (https://github.com/NethServer/nethsecurity-ui/blob/aacbb41dd3115506ebfc4dabbc5d7b0c5e4b0d13/src/stores/standalone/firewall.ts#L184) the value `GUESTS` is still being used. Shouldn't it be changed to `GUEST`?

kevincela (approved):

PR #205, opened 2024-03-05
feat: backup, always show download button

gsanchietti:


gsanchietti:
I can't find quick and pleasant way to display the button.

This is the result:
image


Let's wait @andre8244 for a better fix.

gsanchietti:
It seems that the action is not clean enough.
Also you do the same think by clicking on "Run backup" and then download it.

PR #229, opened 2024-04-02
Controller: add users page

kevincela:
- Add users management page for controller

### DIFF
--- a/public/i18n/en/translation.json
+++ b/public/i18n/en/translation.json
@@ -1844,7 +1844,8 @@
       "delete_user": "Delete user",
       "delete_user_message": "You are about to delete user '{name}'",
       "user_deleted": "User deleted",
-      "unchanged": "Unchanged"
+      "unchanged": "Unchanged",
+      "edit_user": "Edit user"
     }
   },
   "notifications": {
--- a/public/i18n/en/translation.json
+++ b/public/i18n/en/translation.json
@@ -1829,7 +1829,7 @@
     },
     "users": {
       "title": "Users",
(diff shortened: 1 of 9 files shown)


public/i18n/en/translation.json
```diff
@@ -1826,6 +1826,26 @@
       "password_suggestion_4": "At least one number (0-9)",
       "password_suggestion_5": "At least one special character (!{'@'}{'#'}^&_?-)",
       "current_password": "Current password"
+    },
+    "users": {
+      "title": "Users",
+      "description": "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Vivamus lobortis interdum sollicitudin. Proin eget accumsan massa.",
```
gsanchietti, line 1832:
```suggestion
      "description": "The users listed on this page are granted full control over all firewall units. Exclusive permissions to create, modify, or delete user accounts are reserved solely for the administrator."",
```

What do you think @andre8244 ?

src/stores/controller/controllerLogin.ts
```diff
@@ -14,19 +14,22 @@ import { useThemeStore } from '../theme'
...
   })
+  const isAdmin = computed(() => role.value === 'admin')
 
   const loadUserFromStorage = () => {
     const loginInfo = getJsonFromStorage('controllerLoginInfo')
 
     if (loginInfo) {
       username.value = loginInfo.username
       token.value = loginInfo.token
+      role.value = JSON.parse(atob(loginInfo.token.split('.')[1])).role
```
andre8244, line 32: We already have imported `jwt-decode` package to extract token contents. Don't know if it's more reliable than decoding the token manually as you did here, but good to know.

andre8244 (approved):

PR #244, opened 2024-04-16
Controller: handle expiration and refresh of units token

andre8244:
- Open unit: handle expiration and refresh of unit token
- Refresh token of controller user every 30 mins

### DIFF
--- a/src/stores/controller/controllerLogin.ts
+++ b/src/stores/controller/controllerLogin.ts
@@ -43,6 +43,7 @@ export const useLoginStore = defineStore('controllerLogin', () => {
     const loginInfo = {
       username: user,
       token: jwtToken
+      // tokenRefreshedTime: refreshedTime //// need to implement refresh on controller UI
     }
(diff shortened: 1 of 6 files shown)


src/stores/controller/controllerLogin.ts
```diff
@@ -12,9 +12,13 @@ import { getControllerRoutePrefix } from '@/lib/router'
 import { useThemeStore } from '../theme'
 
 export const useLoginStore = defineStore('controllerLogin', () => {
+  const TOKEN_REFRESH_INTERVAL = 1000 * 60 * 30 // half an hour
```
Tbaile, line 15: This is bugging me like the one in [`src/stores/controller/units.ts`](https://github.com/NethServer/nethsecurity-ui/pull/243/files#diff-464476155c61f03f78cf041cbcc1df9a28a98d037290e8f95b9024d53a13be97R50)

Tbaile (approved):

PR #247, opened 2024-04-18
Fix axios error when committing changes

andre8244:
Some browsers (e.g. Firefox) detect an axios error if the page is reloaded just after a POST request.
This happens e.g. when committing UCI changes: the UI waits for the API call to complete and reload the page immediately after.
To mitigate this issue this PR just adds a `setTimeout` before the page reload.

Other changes:
- Bump axios version


Tbaile (approved):
Weird, considering that you wait for the post to return, seems good

PR #251, opened 2024-04-23
Autoreload VPN pages

andre8244:
Automatically reload data inside VPN pages every 10 seconds:
- OpenVPN Road Warrior
- OpenVPN tunnel
- IPsec tunnel

PR #255, opened 2024-04-26
fix: ipsec tunnel, hide skeleton after first load

gsanchietti:
Previously, the skeleton was visible until the first execution of the interval

Related to https://github.com/NethServer/nethsecurity-ui/pull/251


cotosso:
I confirm that now it works as expected.

andre8244 (approved):
Oops :smile:

PR #256, opened 2024-04-26
route add blob gateway updated

cotosso:
blob message changed to cover missing  gateway option

https://github.com/NethServer/nethsecurity/issues/452

### DIFF
--- a/public/i18n/en/translation.json
+++ b/public/i18n/en/translation.json
@@ -1003,7 +1003,7 @@
       "route_network_address": "Network address",
       "route_network_address_cidr": "Network address (CIDR notation)",
       "route_gateway": "Gateway",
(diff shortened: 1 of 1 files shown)


gsanchietti (approved):

public/i18n/en/translation.json
```diff
@@ -1003,7 +1003,7 @@
       "route_network_address": "Network address",
       "route_network_address_cidr": "Network address (CIDR notation)",
       "route_gateway": "Gateway",
-      "route_gateway_helper": "Enter the gateway IP address. The gateway address should be reacheable by the firewall.",
+      "route_gateway_helper": "Enter the gateway IP address. If omitted a link-scope routes will be created. If set to 0.0.0.0 no gateway will be specified for the route.",
```
andre8244, line 1006:
```suggestion
      "route_gateway_helper": "Enter the gateway IP address. If omitted, a link-scope route will be created. If set to 0.0.0.0, no gateway will be specified for the route.",
```

PR #257, opened 2024-04-26
feat: dhcp, expose force option

gsanchietti:
See:
- https://github.com/NethServer/nethsecurity/issues/454
- https://github.com/NethServer/nethsecurity/pull/463

Screenshots:
image

image

### DIFF
--- a/public/i18n/en/translation.json
+++ b/public/i18n/en/translation.json
@@ -978,7 +978,7 @@
       "filter_change_suggestion": "Try changing filter value",
       "dhcp_option_combobox_placeholder": "Search key",
       "force_start": "Force DHCP server start",
-      "force_start_tooltip": "When force option is disabled, the DHCP will not start if there is another DHCP in the network. Disable the force option, only if you want to avoid conflicts with other DHCP servers."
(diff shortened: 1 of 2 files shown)


public/i18n/en/translation.json
```diff
@@ -976,7 +976,9 @@
       "interface_not_found": "Interface not found",
       "no_dynamic_lease_found": "No dynamic leases found",
       "filter_change_suggestion": "Try changing filter value",
-      "dhcp_option_combobox_placeholder": "Search key"
+      "dhcp_option_combobox_placeholder": "Search key",
+      "force_start": "Force DHCP server start",
+      "force_start_tooltip": "When force option is disabled, the DHCP will not start if there is another DHCP in the network. Disable the force option, only if you want to avoid conflicts with other DHCP servers."
```
andre8244, line 981: Maybe it' more clear using positive logic:
```suggestion
      "force_start_tooltip": "When force option is enabled, DHCP will start even if there is another DHCP server in the network. Disable force option only if you want to avoid conflicts with other DHCP servers."
```

andre8244 (approved):

PR #264, opened 2024-05-02
fix: style of disabled firewall rules

andre8244:
- Change background color of disabled rules
- Show gray (`secondary`) badges for disabled rules
- Fix position of Edit and kebab buttons
- Fix standard background color of NeTable row for light theme

https://github.com/NethServer/nethsecurity/issues/480

### DIFF
--- a/src/components/standalone/NeTable.vue
+++ b/src/components/standalone/NeTable.vue
@@ -242,7 +242,7 @@ table {
 }
 
 .table-card td {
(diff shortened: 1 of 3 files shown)


Tbaile (commented):
Really don't like the custom color there, is it possible to store it somewhere? 
If not (or is too much of a pain), go with the merge

andre8244:
> Really don't like the custom color there, is it possible to store it somewhere? If not (or is too much of a pain), go with the merge

I agree with you, but this might be the only time we use it. I propose to store it inside `tailwind.config.js` if/when we will use it again

PR #274, opened 2024-05-13
refactor: import new NeTable and other enhancements

andre8244:
- Controller units: import NeTable from vue-components
- Controller users: import NeTable from vue-components
- Login page: show "Session has expired" when logging out because of token expiration
- Zones and policies: minor fixes
- Dashboard: fix update interval shown at the top of the page
- Replace "&" with "and" in i18n strings
- Clean up code

### DIFF
--- a/src/ControllerApp.vue
+++ b/src/ControllerApp.vue
@@ -123,8 +123,11 @@ function configureAxios() {
             notificationsStore.createNotificationFromAxiosError(error)
           }
         } else {
-          console.warn('[interceptor]', 'Detected error 401, logout')
-          loginStore.logout()
+          if (error.response?.data?.message === 'Token is expired') {
+            console.warn('[interceptor]', 'Detected error 401, logout')
+            loginStore.isSessionExpired = true
+            loginStore.logout()
+          }
         }
       } else {
         // show error notification only if error is not caused from cancellation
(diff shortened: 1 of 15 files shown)


src/ControllerApp.vue
```diff
@@ -123,8 +123,11 @@ function configureAxios() {
             notificationsStore.createNotificationFromAxiosError(error)
           }
         } else {
-          console.warn('[interceptor]', 'Detected error 401, logout')
-          loginStore.logout()
+          if (error.response?.data?.message === 'Token is expired') {
```
Tbaile, line 126: Why this change?
Is there any other reason why a 401 might be returned?

src/stores/controller/controllerLogin.ts
Tbaile, line 1: Shouldn't `isSessionExpired` be a computed property?

How we checked before session expired?

In addition, we could send the user directly to the login page using [Navigation Guards](https://router.vuejs.org/guide/advanced/navigation-guards.html#Per-Route-Guard)

Tbaile (changes requested):

src/ControllerApp.vue
andre8244: > Is there any other reason why a 401 might be returned?

Yes, on wrong username or password :)

Before this change the logout was triggered even when entering a wrong password

src/stores/controller/controllerLogin.ts
andre8244: > Shouldn't isSessionExpired be a computed property?

Don't think so, computed on what?

> How we checked before session expired?

We didn't show "Session has expired" before this PR

> In addition, we could send the user directly to the login page using [Navigation Guards](https://router.vuejs.org/guide/advanced/navigation-guards.html#Per-Route-Guard)

This might be a little improvement but it requires analysis and code changes. I wouldn't do it if current implementation already gets the job done

Tbaile (approved):

PR #275, opened 2024-05-13
fix: slow down update interval of dashboard

andre8244:
Slow down to 20-30 seconds the update interval of dashboard cards


Tbaile (approved):
Hope this solves the issues 🙏

PR #280, opened 2024-05-20
refactor: use NeHeading instead of NeTitle

andre8244:
Use the new NeHeading component instead of NeTitle


andre8244:
CI will succeed after releasing the new version of vue-components

Tbaile (approved):

PR #290, opened 2024-06-07
Refactor: Trim whitespace in input fields

stephdl:
This pull request refactors multiple components to trim whitespace in input fields.

The affected components are RuleEditor.vue, RuleCreator.vue, PolicyEditor.vue, PolicyCreator.vue, and MultiWanGeneralSettings.vue.

The whitespace trimming is done using the v-model.trim directive.

This improves the user experience by automatically removing leading and trailing whitespace from input values.

https://github.com/NethServer/nethsecurity/issues/580

### DIFF
--- a/src/components/standalone/multi-wan/MultiWanGeneralSettings.vue
+++ b/src/components/standalone/multi-wan/MultiWanGeneralSettings.vue
@@ -178,7 +178,7 @@ onMounted(() => {
           <div v-for="index in data.track_ip.keys()" :key="index" class="flex items-center">
             <div class="grow">
               <NeTextInput
-                v-model="data.track_ip[index]"
+                v-model.trim="data.track_ip[index]"
                 :disabled="sending"
                 :invalid-message="messageBag.get(`track_ip.${index}`)?.[0]"
               />
--- a/src/components/standalone/multi-wan/PolicyCreator.vue
+++ b/src/components/standalone/multi-wan/PolicyCreator.vue
@@ -150,7 +150,7 @@ function create() {
     <div v-else class="space-y-8">
       <NeTextInput
         ref="labelElement"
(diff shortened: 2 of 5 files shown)


andre8244:
> Isn't better to trim these values in the backend? If someone calls the CLI with a space the issue will persist

You are right but I fear most (all?) of our APIs don't do any trim. A UI trim is the first layer of defense against bad user input and can probably cover 99% of the scenarios. That being said, IMO it would be a good idea to apply input trim in out future APIs.

andre8244 (approved):

gsanchietti (commented):
Does mwan3 fail if these fields contains a space?

stephdl:
image

stephdl:
for the policy, the label got a space in its name, not sure it is good : `"name": "ns_foo2", "label": "foo2 "`

```
root@NethSec:~#  echo '{}' | /usr/libexec/rpcd/ns.mwan call index_policies
{"values": [{"name": "ns_default", "label": "Default", "members": {"10": [{"name": "ns_wan_M10_W100", "interface": "wan", "metric": "10", "weight": "100", "status": "online"}, {"name": "ns_wan2_M10_W100", "interface": "wan2", "metric": "10", "weight": "100", "status": "online"}]}, "type": "balance"}, {"name": "ns_foo2", "label": "foo2 ", "members": {"10": [{"name": "ns_wan_M10_W100", "interface": "wan", "metric": "10", "weight": "100", "status": "online"}, {"name": "ns_wan2_M10_W100", "interface": "wan2", "metric": "10", "weight": "100", "status": "online"}]}, "type": "balance"}]}
```

gsanchietti (approved):
OK, let's merge it but it seems to me that is not strictly needed.

PR #308, opened 2024-06-28
fix: i18n format of user bind dn tooltip message

andre8244:
Add remote user database: fix i18n message format of **Custom user bind DN** field tooltip. The `@` character is special and requires this format: `{'@'}`


gsanchietti (approved):
:scream:

PR #314, opened 2024-07-05
Custom user bind dn tooltip

cotosso:
updated tooltip for custom user bind dn

### DIFF
--- a/public/i18n/en/translation.json
+++ b/public/i18n/en/translation.json
@@ -1668,7 +1668,7 @@
       "bind_password": "Bind password",
       "bind_password_tooltip": "Password for Bind DN, used to authenticate the connection to the LDAP server",
       "user_bind_dn": "Custom user bind DN",
-      "user_bind_dn_tooltip": "Distinguished Name (DN) used to bind to the LDAP directory. This overrides the calculated user bind DN. Typical configuration is: For Active Directory: '%u@domain', which appends '@domain' to the username. For OpenLDAP: leave this field blank",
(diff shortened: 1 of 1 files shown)


public/i18n/en/translation.json
```diff
@@ -1668,7 +1668,7 @@
       "bind_password": "Bind password",
       "bind_password_tooltip": "Password for Bind DN, used to authenticate the connection to the LDAP server",
       "user_bind_dn": "Custom user bind DN",
-      "user_bind_dn_tooltip": "Distinguished Name (DN) used to bind to the LDAP directory. This overrides the calculated user bind DN. Use it when authenticating against an Active Directory. It can contain a '%u' variable that will be expanded to the user name. Suggestion: use '%u{'@'}secretexample.com' or leave blank for NethServer OpenLDAP",
+      "user_bind_dn_tooltip": "Distinguished Name (DN) used to bind to the LDAP directory. This overrides the calculated user bind DN. Typical configuration is: For Active Directory: '%u@domain', which appends '@domain' to the username. For OpenLDAP: leave this field blank",
```
andre8244, line 1671:
```suggestion
      "user_bind_dn_tooltip": "Distinguished Name (DN) used to bind to the LDAP directory. This overrides the calculated user bind DN. Active Directory usually uses '%u@domain', which appends '@domain' to the username. For OpenLDAP, leave this field blank",
```

I suggest to remove the double column (:) from the sentence

cotosso: I accepted the suggestion, but changed "usually uses" in "typically uses"

PR #316, opened 2024-07-16
fix(i18n): change label for blocklist subscription

gsanchietti:
Blocklists are now available both for Enterprise and Community subscriptions


edospadoni (approved):

andre8244 (approved):

andre8244 (changes requested):
Need to add a parameter to i18n string

PR #317, opened 2024-07-17
fix(interfaces): updating maxLen due to recent mwan issues

Tbaile:
Ref:
- https://github.com/NethServer/nethsecurity/issues/650

### DIFF
--- a/src/components/standalone/interfaces_and_devices/ConfigureDeviceDrawer.vue
+++ b/src/components/standalone/interfaces_and_devices/ConfigureDeviceDrawer.vue
@@ -652,8 +652,8 @@ function validate() {
       }
     } else {
       // check syntax
-      {
-        let maxLen = 15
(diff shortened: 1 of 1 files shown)


gsanchietti (approved):

andre8244 (approved):

gsanchietti:
Is it possible to change the name of the interface when editing?

If the answer is yes, previous version seems better to me: when changing the name, the user is forced to reduce the length and avoid errors.

Tbaile:
> Is it possible to change the name of the interface when editing?
> 
> If the answer is yes, previous version seems better to me: when changing the name, the user is forced to reduce the length and avoid errors.

Nope, it can be done only by destroying and re-creating the interface, it's impossible to rename it after it has been configured, this is why the change :cry:

gsanchietti (approved):

PR #323, opened 2024-07-24
feat(mwan): added object support to mwan

Tbaile:
Adding support for objects in MWAN rules

Ref:
- Issue: https://github.com/NethServer/nethsecurity/issues/671
- Merge after: https://github.com/NethServer/nethsecurity/pull/674

---

Images:
image
image
image

### DIFF
--- a/src/components/standalone/users_objects/DomainSets.vue
+++ b/src/components/standalone/users_objects/DomainSets.vue
@@ -15,6 +15,7 @@ import DeleteModal from '@/components/DeleteModal.vue'
 import CannotDeleteObjectModal from './CannotDeleteObjectModal.vue'
 import ObjectUsagesModal from './ObjectUsagesModal.vue'
 import { useDomainSets, type DomainSet } from '@/composables/useDomainSets'
+import { useObjectStore } from '@/stores/standalone/objects'
 
 const { t } = useI18n()
 const uciChangesStore = useUciPendingChangesStore()
@@ -26,6 +27,8 @@ const {
   errorListDomainSetsDetails,
   searchStringInDomainSet
 } = useDomainSets()
+
+const objects = useObjectStore()
 
 const currentDomainSet = ref<DomainSet | undefined>(undefined)
 const currentDomainSetName = ref('')
@@ -54,6 +57,7 @@ onMounted(() => {
 
 function loadData() {
   listDomainSets()
+  objects.load()
   uciChangesStore.getChanges()
 }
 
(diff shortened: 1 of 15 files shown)


src/components/standalone/users_objects/ObjectUsages.vue
```diff
@@ -85,7 +85,8 @@ function getManagementPageLabel(category: string) {
       return t('common.go_to_page', { page: t('standalone.firewall_rules.title') })
     case 'objects':
       return t('common.go_to_page', { page: t('standalone.objects.title') })
-    //// add other pages
+    case 'mwan3':
```
andre8244, line 88: I'd leave `//// add other pages`, we'll need it for port forward

src/components/standalone/users_objects/ObjectUsages.vue
```diff
@@ -98,7 +99,9 @@ function goToManagementPage(subtype: string) {
       break
     case 'objects':
       path = 'users-objects/objects'
-    //// add other pages
+      break
```
andre8244, line 102: Same as above, I'd leave `//// add other pages`, we'll need it for port forward

andre8244 (commented):
- Objects page is left untouched. For the sake of maintenance, if we adopt this new implementation I suggest to use it everywhere
- Object are never reloaded. I'd do that in Objects page:
  - on page load
  - on object creation/modification

Tbaile:
> * Objects page is left untouched. For the sake of maintenance, if we adopt this new implementation I suggest to use it everywhere

This will drastically add more time to fix due to refactoring everything, not sure if it's better to leave the deprecation warnings and then refactor it when additional changes are needed. Otherwise more (probably much) work is needed.

> * Object are never reloaded. I'd do that in Objects page:
>   
>   * on page load
>   * on object creation/modification

Once the store loads into the app, the fetch is automatic. Adding refresh when creating/editing now.

src/components/standalone/users_objects/DomainSets.vue
```diff
@@ -54,6 +57,7 @@ onMounted(() => {
 
 function loadData() {
   listDomainSets()
+  objects.load()
```
andre8244, line 60: I guess these changes needs to be done in `HostSets.vue` too

Tbaile: Whops, forgot to add them to the commit, coming right up

andre8244 (approved):

PR #324, opened 2024-07-25
fix: always show zone in firewall rules (if not "Any")

andre8244:
See:
- https://github.com/NethServer/nethsecurity/issues/677

Other changes:
- Add vscode extension to devcontainer
- Fix typo in comment

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -28,7 +28,8 @@
         "dbaeumer.vscode-eslint",
         "GitHub.vscode-github-actions",
         "ms-azuretools.vscode-docker",
-        "timonwong.shellcheck"
+        "timonwong.shellcheck",
+        "Vue.volar"
       ]
     }
   }
--- a/src/components/standalone/firewall/rules/SourceOrDestinationRuleColumn.vue
+++ b/src/components/standalone/firewall/rules/SourceOrDestinationRuleColumn.vue
@@ -160,7 +160,7 @@ const zone = computed(() => {
         </p>
       </template>
     </template>
-    <template v-else-if="zone === '*'">
+    <template v-else-if="zone == '*'">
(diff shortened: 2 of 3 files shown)


.devcontainer/devcontainer.json
Tbaile, line 1: How this wasn't in it already

src/components/standalone/firewall/rules/SourceOrDestinationRuleColumn.vue
```diff
@@ -160,20 +160,17 @@ const zone = computed(() => {
         </p>
       </template>
     </template>
-    <template v-else>
+    <template v-else-if="zone === '*'">
```
Tbaile, line 163: Since zone is always a string

```suggestion
    <template v-else-if="zone == '*'">
```

src/components/standalone/firewall/rules/SourceOrDestinationRuleColumn.vue
```diff
@@ -160,20 +160,17 @@ const zone = computed(() => {
...
-            :text="zone.toUpperCase()"
-            :kind="enabled ? 'custom' : 'secondary'"
-            :customColorClasses="getZoneColorClasses(zone)"
-          />
-        </span>
+        {{ t('common.any') }}
       </div>
     </template>
+    <NeBadge
+      v-if="zone && zone !== '*'"
```
Tbaile, line 170:
```suggestion
      v-if="zone && zone != '*'"
```

.devcontainer/devcontainer.json
andre8244: At first the requirement was to not show the zone if some addresses/objects were specified in the rule

Tbaile: Was talking about the `"Vue.volar"` extension that was missing :sweat_smile:

Tbaile (approved):

PR #326, opened 2024-07-26
fix: add openvpn dynamic ip range

andre8244:
See:
- https://github.com/NethServer/nethsecurity/issues/678

Other changes:
- Set focus on the first invalid field while creating the OpenVPN RW server
- Solve linter warnings
- Remove commented code


andre8244:
Need to replace a "Lorem ipsum" for this tooltip, any ideas @gsanchietti ?

image

Tbaile (approved):

PR #333, opened 2024-08-06
Improve mwan rule form

gsanchietti:
Changes:
- clean form after save
- mimic firewall rules form for a better user experience

Issue: NethServer/nethsecurity#671

### DIFF
--- a/public/i18n/en/translation.json
+++ b/public/i18n/en/translation.json
@@ -970,16 +970,14 @@
       "protocol_udp": "UDP",
       "protocol_icmp": "ICMP",
       "protocol_all": "All",
-      "address_options": {
-        "any": "Any",
-        "address": "Address",
-        "object": "Object"
-      },
+      "address_options_address": "Enter an address",
+      "address_options_object": "Select an object",
+      "address_options_any_src": "Any source address",
+      "address_options_any_dst": "Any destination address",
       "source_type": "Source type",
       "destination_type": "Destination type",
       "source_object": "Source object",
       "enter_address": "Enter an address",
-      "select_an_object": "Select an object",
       "destination_object": "Destination object"
     },
     "dns_dhcp": {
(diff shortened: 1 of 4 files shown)


public/i18n/en/translation.json
```diff
@@ -970,16 +970,14 @@
       "protocol_udp": "UDP",
       "protocol_icmp": "ICMP",
       "protocol_all": "All",
-      "address_options": {
-        "any": "Any",
-        "address": "Address",
-        "object": "Object"
-      },
+      "address_options_address": "Enter one address",
```
andre8244, line 973:
```suggestion
      "address_options_address": "Enter an address",
```

src/composables/useRuleForm.ts
```diff
@@ -101,14 +101,20 @@ export function useRuleForm(policies: Ref<Policy[]>, rule?: Ref<Rule | undefined
...
-  const srcType = ref<AddressType>('any')
-  const dstType = ref<AddressType>('any')
+  const dstAddressOptions: { id: AddressType; label: string }[] = [
+    { id: 'address', label: t('standalone.multi_wan.address_options_address') },
+    { id: 'any', label: t('standalone.multi_wan.address_options_any_dst') },
+    { id: 'object', label: t('standalone.multi_wan.address_options_object') }
+  ]
+
+  const srcType = ref<AddressType>('address')
+  const dstType = ref<AddressType>('address')
```
andre8244, line 117: Are we changing the default value from `any` to `address`?

andre8244 (approved):
If we want to change the default value of source and destination, LGTM

src/composables/useRuleForm.ts
gsanchietti: Yes, now it's the same of firewall rules

public/i18n/en/translation.json
gsanchietti: Fixed

Tbaile (approved):

PR #335, opened 2024-08-27
fix: nat rules, add 0.0.0.0/0 as destination address

andre8244:
Ref:
- https://github.com/NethServer/nethsecurity/issues/720

### DIFF
--- a/src/components/standalone/firewall/nat/CreateOrEditNatRuleDrawer.vue
+++ b/src/components/standalone/firewall/nat/CreateOrEditNatRuleDrawer.vue
@@ -94,12 +94,6 @@ const actionOptions = ref([
   }
 ])
 
-const anyAddress = ref({
-  id: '0.0.0.0/0',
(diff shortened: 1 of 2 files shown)


src/components/standalone/firewall/nat/CreateOrEditNatRuleDrawer.vue
```diff
@@ -93,6 +94,12 @@ const actionOptions = ref([
   }
 ])
 
+const anyAddress = ref({
+  id: '0.0.0.0/0',
+  label: '0.0.0.0/0',
+  description: t('standalone.nat.any_address')
+})
```
Tbaile, line 101: This can be a `dict` without ref wrapping? Standing by the code, it appears that afterwards you'll have a mixed `destinationAddressOptions` which contains both a `ref` of this and standard `dict`s

Tbaile (approved):

PR #349, opened 2024-09-05
Controller version awareness

Tbaile:
Implementing what said in [this issue](https://github.com/NethServer/nethsecurity/issues/721#issuecomment-2329479009).

### DIFF
--- a/public/i18n/en/translation.json
+++ b/public/i18n/en/translation.json
@@ -2045,7 +2045,12 @@
       "num_units_left_tooltip": "You have configured {num} of {max} units. Upgrade to a subscription plan to manage more units.",
       "sync_unit_info": "Sync unit info",
       "syncing_data": "Syncing data...",
-      "syncing_data_description": "Unit data hasn't been synced yet. You can do it now by clicking the three dots menu and then '@:controller.units.sync_unit_info '."
+      "syncing_data_description": "Unit data hasn't been synced yet. You can do it now by clicking the three dots menu and then '@:controller.units.sync_unit_info '.",
+      "cannot_open_unit": "Cannot open unit",
+      "cannot_open_unit_description": "The unit {name} must be updated before opening it. Access the unit UI directly and go to System > Updates page.",
+      "open_anyway": "Open anyway",
+      "warning_open_unit": "Controller is not up-to-date",
+      "warning_open_unit_description": "It appears that the controller is not up-to-date. It's recommended to update the controller instance on your NethServer 8 cluster before opening this unit to avoid compatibility issues."
     },
     "unit_terminal": {
       "name_unit_terminal": "{name} unit terminal",
--- a/src/components/controller/units/UnitsTable.vue
+++ b/src/components/controller/units/UnitsTable.vue
@@ -6,32 +6,34 @@
 <script setup lang="ts">
 import { useI18n } from 'vue-i18n'
 import {
+  byteFormat1024,
+  deleteFromStorage,
+  formatDateLoc,
+  getAxiosErrorMessage,
+  getPreference,
   NeButton,
   NeDropdown,
   NeInlineNotification,
   NeLink,
-  NeTooltip,
+  NeModal,
+  NePaginator,
   NeTable,
+  NeTableBody,
+  NeTableCell,
   NeTableHead,
   NeTableHeadCell,
-  NeTableBody,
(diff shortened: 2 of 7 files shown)


public/i18n/en/translation.json
```diff
@@ -2045,7 +2045,12 @@
       "num_units_left_tooltip": "You have configured {num} of {max} units. Upgrade to a subscription plan to manage more units.",
       "sync_unit_info": "Sync unit info",
       "syncing_data": "Syncing data...",
-      "syncing_data_description": "Unit data hasn't been synced yet. You can do it now by clicking the three dots menu and then '@:controller.units.sync_unit_info '."
+      "syncing_data_description": "Unit data hasn't been synced yet. You can do it now by clicking the three dots menu and then '@:controller.units.sync_unit_info '.",
+      "cannot_open_unit": "Cannot open unit",
+      "cannot_open_unit_description": "The unit {name} must be updated before opening it. Access the unit UI directly and go to System > Updates page.",
+      "open_anyway": "Open anyway",
+      "warning_open_unit": "Unit is not up-to-date",
```
andre8244, line 2052:
```suggestion
      "warning_open_unit": "Controller is not up-to-date",
```

public/i18n/en/translation.json
```diff
@@ -2045,7 +2045,12 @@
...
+      "warning_open_unit_description": "It appears that the unit is not up-to-date with the latest version. It's recommended to update the Controller before opening this unit to avoid compatibility issues."
```
andre8244, line 2053:
```suggestion
      "warning_open_unit_description": "It appears that the controller is not up-to-date. It's recommended to update the controller instance on your NethServer 8 cluster before opening this unit to avoid compatibility issues."
```

src/components/controller/units/UnitsTable.vue
```diff
@@ -63,16 +68,39 @@ onMounted(() => {
...
-    await unitsStore.checkUnitToken(unitId)
-    const routeData = router.resolve({ path: `/controller/manage/${unitId}/dashboard` })
-    window.open(routeData.href, '_blank')
+    await unitsStore.getUnitInfo(currentUnit.value.id)
+    await unitsStore.getUnits()
+    // Find the now updated unit, as the currentUnit might have changed
+    currentUnit.value = unitsStore.units.find((u) => u.id == unit.id)!
+    // Logic on which message is shown is inside the ObsoleteApiModal component
+    let version = currentUnit.value?.info.api_version
+    if (version == '') {
```
andre8244, line 83: When does this happen? Can `version` also be `null`  or `undefined`?

src/components/controller/units/UnitsTable.vue
```diff
@@ -63,16 +68,39 @@ onMounted(() => {
...
+      version = '0.0.0'
+    }
+    if (versionCheck && !satisfies(version, REQUIRED_API_VERSION)) {
+      if (outside(version, REQUIRED_API_VERSION, '>')) {
+        showGreaterApiModal.value = true
```
andre8244, line 88: I guess `showObsoleteApiModal.value = false` is missing here. Maybe the best solution is to set:
```
showGreaterApiModal.value = false
showObsoleteApiModal.value = false
```
at the start of this function, i.e. at line 75

andre8244 (changes requested):

src/components/controller/units/UnitsTable.vue
Tbaile: It can if you're browsing a current stable unit with the latest controller, the `api_version` field is empty, while `currentUnit.value?` is `Unit | undefined` by nature

src/components/controller/units/UnitsTable.vue
```diff
@@ -63,16 +68,39 @@ onMounted(() => {
...
+      } else {
+        showObsoleteApiModal.value = true
```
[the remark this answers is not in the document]
Tbaile: Same as above

src/components/controller/units/UnitsTable.vue
Tbaile: Now that i think about it, the setting to false of the two flags is not needed, the only flag needed to be set to false is `showGreaterApiModal` since it's possible that is set to true due to the warning modal. Your call boss!

andre8244 (approved):

PR #352, opened 2024-09-10
feat: nat helpers UI

andre8244:
Ref:
- https://github.com/NethServer/nethsecurity/issues/724

### DIFF
--- a/src/components/standalone/firewall/nat/EditNatHelperDrawer.vue
+++ b/src/components/standalone/firewall/nat/EditNatHelperDrawer.vue
@@ -1,0 +1,265 @@
+<!--
+  Copyright (C) 2024 Nethesis S.r.l.
+  SPDX-License-Identifier: GPL-3.0-or-later
+-->
+
+<script setup lang="ts">
+import { useI18n } from 'vue-i18n'
+import { ref } from 'vue'
+import { MessageBag, validateRequired, type validationOutput } from '@/lib/validation'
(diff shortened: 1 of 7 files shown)


src/components/standalone/firewall/nat/EditNatHelperDrawer.vue
```diff
@@ -0,0 +1,265 @@
...
+  const natHelperName = props.natHelper?.name
+
+  const payload = {
+    name: natHelperName,
+    enabled: enabled.value,
+    // convert params array to object
+    params: params.value.reduce((acc, param) => {
+      acc[param.name] = param.value
+      return acc
+    }, {} as Record<string, string>)
```
Tbaile, line 135: Is this the proper way to use reduce? Isn't a `map` method better?

Tbaile: Definitely not liking it, but okay (confirming that AI won't be able to steal our jobs)

Tbaile (approved):

src/components/standalone/firewall/nat/EditNatHelperDrawer.vue
andre8244: You're right, I wasn't sure either at first :joy: However we have an explanatory human comment above that line: `// convert params array to object` :innocent:

PR #370, opened 2024-09-27
feat(controller): add remote unit update

Tbaile:
Ref:
- https://github.com/NethServer/nethsecurity/issues/800
- https://github.com/NethServer/nethsecurity/pull/801


src/assets/main.css
```diff
@@ -4,6 +4,7 @@
  */
 
 @import '@xterm/xterm/css/xterm.css';
+
```
gsanchietti, line 7: can we cleanup this?

src/components/controller/units/ScheduleUpdateDrawer.vue
```diff
@@ -0,0 +1,230 @@
...
+  --dp-primary-color: rgb(6 182 212 / var(--tw-bg-opacity));
+  --dp-primary-text-color: rgb(3 7 18 / var(--tw-text-opacity));
+  --dp-border-color-hover: var(--dp-primary-color);
+}
+
+.dp__theme_light {
+  --dp-primary-color: rgb(14 116 144 / var(--tw-bg-opacity));
+  --dp-border-color-hover: var(--dp-primary-color);
+}
+</style>
```
gsanchietti, line 230: This has been already used: is it possible to avoid duplication?

src/components/controller/units/UnitsTable.vue
```diff
@@ -42,7 +51,17 @@ const props = defineProps({
   }
 })
 
-const emit = defineEmits(['reloadData', 'openSshModal'])
+// TODO: when dropdown nicely implements icons, remove this with correct imports directly in the component
```
gsanchietti, line 54:
```suggestion
// FIXME: when dropdown nicely implements icons, remove this with correct imports directly in the component
```

Tbaile: Worry not, a TODO will suffice, if we change the type of the icons the whole build will break :smile:

src/components/controller/units/ScheduleUpdateDrawer.vue
Tbaile: Tried again even now, it seems that the external component doesn't like at all tailwind and adding this forces the correct styling for it. I've tried but for now this is all that is possible.

Tbaile: Pushed now a change that wraps all CSS in itself, but this is the best that is possible right now.

gsanchietti (approved):

PR #397, opened 2024-10-09
fix(mwan): add length validation for name

stephdl:
This pull request adds a length validation for the name rule in the `useRuleForm.ts` file. The validation ensures that the name is not more than 12 characters long. This validation is implemented in the `validateLengthNotMore12` function in the `validation.ts` file.

https://github.com/NethServer/nethsecurity/issues/835

### DIFF
--- a/src/composables/useRuleForm.ts
+++ b/src/composables/useRuleForm.ts
@@ -10,7 +10,8 @@ import {
   MessageBag,
   validateIp4OrCidr,
   validatePortRangeForMwan,
-  validateRequired
+  validateRequired,
+  validateUciName
 } from '@/lib/validation'
 import { type ObjectReference, useObjects } from '@/composables/useObjects'
 import { ubusCall } from '@/lib/standalone/ubus'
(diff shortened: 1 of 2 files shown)


Tbaile:
Shall we add a configurable validation rule? Such as "max {n} characters" are available?

gsanchietti:
I'm going to leave the review to @Tbaile because I do not know enough the components.

src/composables/useRuleForm.ts
```diff
@@ -171,6 +172,14 @@ export function useRuleForm(policies: Ref<Policy[]>, rule?: Ref<Rule | undefined
...
       validationErrors.value.set('policy', String(validationCheck.errMessage))
     }
+    validationCheck = validateUciName(name.value, 12)
+    if (!validationCheck.valid) {
+      validationErrors.value.set('name', String(validationCheck.errMessage)),
+        validationErrors.value.set(
+          'num',
+          String((validationCheck.i18Params as { num?: number })?.num)
+        )
+    }
```
Tbaile, line 182: To see how to handle the translation process of these cases in a more clean way: https://github.com/NethServer/nethsecurity-ui/blob/898bc8593bc8cbdb212ff84e604d0adf4af9ad77/src/components/standalone/interfaces_and_devices/ConfigureDeviceDrawer.vue#L663

Tbaile (changes requested):
While thinking about adding a max length rule, it appears that in this case it's correct to use this kind of validation (due to iptables being a psycho regarding the naming schemes). Just a few touches and this can be pushed to the 8.4 roadmap!

Tbaile (changes requested):
For parameters don't return additional value in the MessageBox, use the `i18Params`

Tbaile (approved):

Tbaile:
Will merge after next NethSecurity release

PR #415, opened 2024-10-29
refactor(lang): changing how translation string are fetched

Tbaile:
Due to recurring issues with translation loading due to cache ([example can be found here](https://community.nethserver.org/t/nethsecurity-ui-language-not-found-after-update/24805)), a major overhaul has been made to the translation section of the application.

The following changes include:
- immediate language change in whole application
- usage of standard labels for language specification
- using Vite asset bundling to allow cache busting 
- created custom component that handles all language changes

NOTE: Some settings might need to be changed in Weblate.


tsconfig.app.json
```diff
@@ -1,6 +1,6 @@
 {
   "extends": "@vue/tsconfig/tsconfig.dom.json",
-  "include": ["env.d.ts", "src/**/*", "src/**/*.vue"],
+  "include": ["env.d.ts", "src/**/*", "src/**/*.vue", "src/i18n/**/*.json"],
```
andre8244, line 3: What is this for?

Tbaile: This is telling typescript to be aware of the files, Vite inherits this and loads them automatically (when imported)

andre8244 (approved):

PR #439, opened 2024-11-20
fix(monitoring): fix y scale of ping delivery rate

andre8244:
- Set the maximum value of the y-axis for the ping delivery rate to "100%"
- Enhance the configurability of `TimeLineChart` by introducing an `options` prop

Ref:
- https://github.com/NethServer/nethsecurity/issues/904


Tbaile (approved):
I'd love to see using types coming from ChartJS, they're not documented but they are there :eyes:

PR #447, opened 2024-11-29
refactor(traffic): refactored traffic section

Tbaile:
Refactored Traffic section following the new API coming from https://github.com/NethServer/nethsecurity/pull/940.

Ref: https://github.com/NethServer/nethsecurity/issues/937

PR #446, opened 2024-11-26
refactor(traffic): moving instant traffic to another section

Tbaile:
Closes: https://github.com/NethServer/nethsecurity/issues/936

--- 

Screenshots:

PR #455, opened 2024-12-05
feat(monitoring): instant traffic tab and filtering options 

Tbaile:
Ref:
- https://github.com/NethServer/nethsecurity/issues/900


andre8244 (approved):
If I'm not mistaken this is the combination of #445, #446 and #447, with minor changes to fit everything together

PR #457, opened 2024-12-05
fix(controller): adding space as non-valid character

Tbaile:
Ref:
- https://github.com/NethServer/nethsecurity/issues/952

### DIFF
--- a/src/lib/validation.ts
+++ b/src/lib/validation.ts
@@ -723,6 +723,17 @@ export const validateDNSForwardingServer = (value: string): validationOutput => 
 }
 
 /**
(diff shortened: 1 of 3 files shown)


src/lib/validation.ts
```diff
@@ -722,6 +722,17 @@ export const validateDNSForwardingServer = (value: string): validationOutput =>
   return validResult
 }
 
+/**
+ * Validate if the string doesn't have spaces.
+ * @param value
+ */
+export function validateSpace(value: string): validationOutput {
```
andre8244, line 729: Let's rename to `validateNoSpaces` for uniformity with other functions

PR #467, opened 2024-12-27
feat: added netify interface

stephdl:
Introduce a new monitoring component for Netify Informatics, including necessary routes and translations for integration into the application.

https://github.com/NethServer/nethsecurity/issues/981

https://github.com/user-attachments/assets/514fb61d-4944-41c5-9798-96983b63f68f

https://github.com/user-attachments/assets/bc35f876-c94d-44d2-a8a1-96d9e810660d

https://github.com/user-attachments/assets/18dd855c-1f9e-482a-a1df-79c69e5ee8ad


src/components/standalone/monitoring/NetifyInformaticsContent.vue
Tbaile, line 1: Please stick to the camelCase format for both variables and functions in the JS/TS space.

src/components/standalone/monitoring/NetifyInformaticsContent.vue
```diff
@@ -0,0 +1,188 @@
...
+interface ErrorObject {
+  notificationTitle: string
+  notificationDescription: string
+  uuid: string
+}
+const objError: ErrorObject = {
+  notificationTitle: '',
+  notificationDescription: '',
+  uuid: ''
+}
```
Tbaile, line 48: While I don't like the approach of multiple error handling like this, I can allow it, but it completely loses effectiveness when there are multiple variables for each error, is redundant (and don't assign them like that, while TS allows it, it's really a nasty approach)

src/components/standalone/monitoring/NetifyInformaticsContent.vue
```diff
@@ -0,0 +1,188 @@
...
+
+async function getConfiguration() {
+  try {
+    loading.value = true
+    let getDataConfiguration = await ubusCall('ns.netifyd', 'status', {})
+    if (getDataConfiguration && getDataConfiguration.data) {
+      let configuration = getDataConfiguration.data
+      form.value.status = configuration.enabled
+      form.value.uuid = configuration.uuid
+      loading.value = false
```
Tbaile, line 72: Put this into a "finally" clause, this way the loading always removes itself after the request

src/components/standalone/monitoring/NetifyInformaticsContent.vue
```diff
@@ -0,0 +1,188 @@
...
+    }
+  } catch (exception: any) {
+    isError.value = true
```
Tbaile, line 75: Maybe is better to define the `error` object as nullable, and check for nullable if there's an error set, otherwise we assume state is good

src/components/standalone/monitoring/NetifyInformaticsContent.vue
```diff
@@ -0,0 +1,188 @@
...
+  }, 3000)
+}
+</script>
+
+<template>
+  <div>
+    <FormLayout class="max-w-6xl">
+      <template #description>
+        <i18n-t keypath="standalone.netify_informatics.content_description" tag="p" scope="global">
+        </i18n-t>
```
Tbaile, line 133: Use `t` object in template for every localized string
```suggestion
        <p>{{ t('standalone.netify_informatics.content_description') }}</p
```

src/components/standalone/monitoring/NetifyInformaticsContent.vue
```diff
@@ -0,0 +1,188 @@
...
+        <NeButton kind="tertiary" size="lg" @click="open_dashboard()" class="ml-2 shrink-0">
+          <template #prefix>
+            <FontAwesomeIcon :icon="['fas', 'arrow-up-right-from-square']" aria-hidden="true" />
```
Tbaile, line 136: Import the icon and use it directly in the `:icon` prop, [like here](https://github.com/NethServer/nethsecurity-ui/blob/dbf54e0203a294860a1a3f974ec599bc5aa2d6d2/src/components/controller/users/UsersTable.vue#L86)

src/components/standalone/monitoring/NetifyInformaticsContent.vue
```diff
@@ -0,0 +1,188 @@
...
+            v-model="form.uuid"
+            disabled
+            :invalid-message="error.uuid"
+            :label="t('standalone.netify_informatics.uuid')"
+            ref="uuidRef"
+          >
+          </NeTextInput>
+          <NeButton kind="tertiary" size="sm" @click="copy_uuid" class="ml-2.5 mt-8">
+            <template #prefix>
+              <font-awesome-icon :icon="['fas', 'copy']" class="h-4 w-4" aria-hidden="true" />
```
Tbaile, line 161: Same as above, also:
```suggestion
              <FontAwesomeIcon :icon="['fas', 'copy']" class="h-4 w-4" aria-hidden="true" />
```

Tbaile (changes requested):

src/i18n/en/translation.json
```diff
@@ -446,6 +447,21 @@
...
   },
   "standalone": {
+    "netify_informatics": {
+      "title": "Netify Informatics",
+      "netify_informatics_enable": "Netify Informatics enabled",
+      "netify_informatics_disable": "Netify Informatics disabled",
+      "netify_informatics_enable_description": "Metadata sending is now active. Configuration saved successfully",
+      "netify_informatics_disable_description": "Metadata sending has been disabled. Configuration saved successfully",
+      "content_description": "Netify Informatics is a third-party cloud service that utilizes analytics and AI to convert local DPI metadata obtained from NethSecurity into high-level network intelligence and visibility.",
+      "uuid": "UUID",
```
gsanchietti, line 457:
```suggestion
      "uuid": "Agent UUID",
```

src/i18n/en/translation.json
```diff
@@ -446,6 +447,21 @@
     "please_sign_in_again": "Please sign in again"
   },
   "standalone": {
+    "netify_informatics": {
+      "title": "Netify Informatics",
+      "netify_informatics_enable": "Netify Informatics enabled",
+      "netify_informatics_disable": "Netify Informatics disabled",
+      "netify_informatics_enable_description": "Metadata sending is now active. Configuration saved successfully",
+      "netify_informatics_disable_description": "Metadata sending has been disabled. Configuration saved successfully",
+      "content_description": "Netify Informatics is a third-party cloud service that utilizes analytics and AI to convert local DPI metadata obtained from NethSecurity into high-level network intelligence and visibility.",
```
gsanchietti, line 456:
```suggestion
      "content_description": "Netify Informatics is a third-party cloud service that utilizes analytics and AI to convert local DPI metadata obtained from NethSecurity into high-level network intelligence. Copy the Agent UUID, then access Netify site and proceed with provision agent wizard.",
```

gsanchietti (commented):
Minor label rewording.

src/views/standalone/monitoring/NetifyInformaticsView.vue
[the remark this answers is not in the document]
Tbaile: Agree when there are multiple components, otherwise you can safely add everything in the page (due to the dynamic import of the router, everything will be pulled only when needed :magic_wand:)

src/components/standalone/monitoring/NetifyInformaticsContent.vue
```diff
@@ -0,0 +1,189 @@
...
+            class="ml-2.5 mt-8"
+            v-if="justCopied"
+            :icon="faCheck"
+            :text="t('standalone.netify_informatics.copied_to_clipboard')"
+            kind="secondary"
+            :rounded="false"
+            size="sm"
+          />
+        </div>
+        <NeFormItemLabel>{{ t('standalone.netify_informatics.status') }}</NeFormItemLabel>
```
andre8244, line 189: This label has an incorrect spacing from the toggle. Let's use `topLabel` prop of `NeToggle` component, see https://nethesis.github.io/vue-components/?path=/story/netoggle--with-top-label

src/components/standalone/monitoring/NetifyInformaticsContent.vue
```diff
@@ -0,0 +1,189 @@
...
+  }, 3000)
+}
+</script>
+
+<template>
+  <div>
+    <FormLayout class="max-w-6xl">
+      <template #description>
+        <p>{{ t('standalone.netify_informatics.content_description') }}</p>
+        <NeButton kind="tertiary" size="lg" @click="openDashboard()" class="ml-2 shrink-0">
```
andre8244, line 141: Let's fix vertical spacing and use `NeLink` instead (in general: NeButton is for actions, NeLink is for simple links). See https://github.com/NethServer/nethsecurity-ui/blob/main/src/components/standalone/backup_and_restore/MigrationContent.vue#L39-L44

src/components/standalone/monitoring/NetifyInformaticsContent.vue
```diff
@@ -0,0 +1,189 @@
...
+        <div class="relative flex items-center gap-2">
+          <NeTextInput
+            v-model="form.uuid"
+            disabled
+            :invalid-message="error.uuid"
+            :label="t('standalone.netify_informatics.uuid')"
+            ref="uuidRef"
+          >
+          </NeTextInput>
+          <NeButton kind="tertiary" size="sm" @click="copyUuid" class="ml-2.5 mt-8">
```
andre8244, line 173: "Copied to clipboard" badge should be a tooltip instead. A bit of rework is needed, let's replace this NeButton and the NeBadge with:

```suggestion
          <NeTooltip v-if="justCopied" triggerEvent="mouseenter focus" placement="top-start">
            <template #trigger>
              <NeButton kind="secondary" size="md" @click="copyUuid" class="mb-px ml-2.5">
                <template #prefix>
                  <FontAwesomeIcon :icon="faCopy" class="h-4 w-4" aria-hidden="true" />
                </template>
                {{ t('standalone.netify_informatics.copy') }}
              </NeButton>
            </template>
            <template #content>
              {{ t('standalone.netify_informatics.copied_to_clipboard') }}
            </template>
          </NeTooltip>
          <NeButton v-else kind="secondary" size="md" @click="copyUuid" class="mb-px ml-2.5">
            <template #prefix>
              <FontAwesomeIcon :icon="faCopy" class="h-4 w-4" aria-hidden="true" />
            </template>
            {{ t('standalone.netify_informatics.copy') }}
          </NeButton>
```

src/components/standalone/monitoring/NetifyInformaticsContent.vue
```diff
@@ -0,0 +1,189 @@
...
+      <NeSkeleton v-if="loading" :lines="5" />
+      <NeInlineNotification
+        v-if="errorLoadingConfiguration.notificationTitle"
+        class="my-4"
+        kind="error"
+        :title="errorLoadingConfiguration.notificationTitle"
+        :description="errorLoadingConfiguration.notificationDescription"
+      />
+      <div v-if="!isError && !loading" class="mb-8 flex flex-col gap-y-6">
+        <div class="relative flex items-center gap-2">
```
andre8244, line 164:
```suggestion
        <div class="relative flex items-end gap-2">
```

src/i18n/en/translation.json
```diff
@@ -446,6 +447,21 @@
     "please_sign_in_again": "Please sign in again"
   },
   "standalone": {
+    "netify_informatics": {
+      "title": "Netify Informatics",
+      "netify_informatics_enable": "Netify Informatics enabled",
+      "netify_informatics_disable": "Netify Informatics disabled",
+      "netify_informatics_enable_description": "Metadata sending is now active. Configuration saved successfully",
```
andre8244, line 454:
```suggestion
      "netify_informatics_enable_description": "Metadata sending is now active",
```

src/i18n/en/translation.json
```diff
@@ -446,6 +447,21 @@
...
+      "netify_informatics_disable_description": "Metadata sending has been disabled. Configuration saved successfully",
```
andre8244, line 455:
```suggestion
      "netify_informatics_disable_description": "Metadata sending has been disabled",
```

src/components/standalone/monitoring/NetifyInformaticsContent.vue
```diff
@@ -0,0 +1,189 @@
...
+import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome'
+import { faArrowUpRightFromSquare, faCopy, faCheck } from '@fortawesome/free-solid-svg-icons'
+import { NeButton } from '@nethesis/vue-components'
+import {
+  NeSkeleton,
+  NeFormItemLabel,
+  NeTextInput,
+  getAxiosErrorMessage,
+  NeInlineNotification,
+  NeBadge
```
andre8244, line 21:
```suggestion
  NeTooltip
```

andre8244 (changes requested):

src/components/standalone/monitoring/NetifyInformaticsContent.vue
stephdl: happy to do with another way tommaso, but it is a common way to set to true when we have an error, what do you think @andre8244

Tbaile: Yes, but in this case the error information is decoupled, during coding if `isError` is not set during an error no message is displayed even if the error is thrown and saved. 
A solution can be of the following:
```ts
const error: Ref<ErrorObject?> = ref()
```
This tells the code that the error can be `undefined` and you can check for it

src/components/standalone/monitoring/NetifyInformaticsContent.vue
```diff
@@ -0,0 +1,200 @@
...
+
+interface Form {
+  status: boolean
+  uuid: string
+}
+
+const form = ref<Form>({
+  status: false,
+  uuid: ''
+})
```
Tbaile, line 37: Nesting is not needed, just adds overhead

```suggestion
const status = ref(false)
cost uuid = ref('')
```

src/components/standalone/monitoring/NetifyInformaticsContent.vue
```diff
@@ -0,0 +1,200 @@
...
+
+const objError = {
+  notificationTitle: '',
+  notificationDescription: '',
+  uuid: ''
+}
+
+const error = ref({ ...objError })
+const errorLoadingConfiguration = ref({ ...objError })
+const errorSaving = ref({ ...objError })
```
Tbaile, line 47: This part is really hard to understand

`error` is effectively not used, if not for line 162, however the value is never set, so what is this for?
`errorLoadingConfiguration` and `errorSaving` are basically the same, there's no difference between them, they can be the same value

src/components/standalone/monitoring/NetifyInformaticsContent.vue
```diff
@@ -0,0 +1,200 @@
...
+        kind="error"
+        :title="errorSaving.notificationTitle"
+        :description="errorSaving.notificationDescription"
+      />
+      <div v-if="!isError && !loading" class="mb-8 flex flex-col gap-y-6">
+        <div class="relative flex items-end gap-2">
+          <NeTextInput
+            v-model="form.uuid"
+            disabled
+            :invalid-message="error.uuid"
```
Tbaile, line 162: A read-only value, can be invalid?

src/components/standalone/monitoring/NetifyInformaticsContent.vue
```diff
@@ -0,0 +1,200 @@
...
+          <NeButton v-else kind="secondary" size="md" @click="copyUuid" class="mb-px ml-2.5">
+            <template #prefix>
+              <FontAwesomeIcon :icon="faCopy" class="h-4 w-4" aria-hidden="true" />
+            </template>
+            {{ t('standalone.netify_informatics.copy') }}
+          </NeButton>
+        </div>
+        <NeToggle
+          v-model="form.status"
+          :topLabel="t('standalone.netify_informatics.status')"
```
Tbaile, line 189:
```suggestion
          :top-label="t('standalone.netify_informatics.status')"
```

src/components/standalone/monitoring/NetifyInformaticsContent.vue
andre8244: Let's find the quickest solution, but IMHO there are too many variables and typescript types for a simple form with only a toggle :confused: If you got stuck I suggest to use the same variables pattern of other pages (e.g. NatHelpers.vue is a recent and simple page)

src/components/standalone/monitoring/NetifyInformaticsContent.vue
```diff
@@ -0,0 +1,200 @@
...
+          @update:model-value="handleSaveAutomaticToggle"
```
Tbaile, line 190: This is dangerous, you're overriding the [vue default compilation](https://vuejs.org/guide/components/v-model.html#under-the-hood).

I'd suggest the usage of `watch` instead, if you used it before and the trigger was not working, could be cause of the missing [deep directive](https://vuejs.org/guide/essentials/watchers.html#deep-watchers), which if you simplify the `refs` this is not needed anymore.

Tbaile (changes requested):

gsanchietti (commented):
Labels are good for me.

Tbaile (dismissed):
Remember to squash and remove the body of the commit, title is fine the one on the PR

andre8244 (dismissed):
Ok, just a minor style fix

andre8244 (approved):

gsanchietti (approved):

PR #468, opened 2025-01-03
feat(firewall rule): highlight rules on non-existent zones

stephdl:
Introduce a true_zone property to the FirewallRule interface, improve display logic to account for zone existence, and add new translations for inactive and disabled firewall rules.

https://github.com/NethServer/nethsecurity/issues/899

https://github.com/user-attachments/assets/463d96b5-78b1-4184-bf6c-61302b7d64f9

https://github.com/user-attachments/assets/f6fd4ac7-e271-4161-a8af-ab6d86b16945

### DIFF
--- a/src/components/standalone/firewall/rules/CreateOrEditFirewallRuleDrawer.vue
+++ b/src/components/standalone/firewall/rules/CreateOrEditFirewallRuleDrawer.vue
@@ -720,8 +720,7 @@ async function saveRule() {
       return tag.id
     }),
     log: isLoggingEnabled.value,
-    system_rule: props.currentRule?.system_rule || false,
-    active_zone: props.currentRule?.active_zone || false
+    system_rule: props.currentRule?.system_rule || false
   }
 
   if (isEditingRule.value && props.currentRule) {
--- a/src/components/standalone/firewall/rules/CreateOrEditFirewallRuleDrawer.vue
+++ b/src/components/standalone/firewall/rules/CreateOrEditFirewallRuleDrawer.vue
@@ -720,7 +720,8 @@ async function saveRule() {
       return tag.id
     }),
     log: isLoggingEnabled.value,
-    system_rule: props.currentRule?.system_rule || false
+    system_rule: props.currentRule?.system_rule || false,
+    true_zone: props.currentRule?.true_zone || false
   }
 
   if (isEditingRule.value && props.currentRule) {
--- a/src/components/standalone/firewall/rules/CreateOrEditFirewallRuleDrawer.vue
+++ b/src/components/standalone/firewall/rules/CreateOrEditFirewallRuleDrawer.vue
@@ -721,7 +721,7 @@ async function saveRule() {
     }),
     log: isLoggingEnabled.value,
     system_rule: props.currentRule?.system_rule || false,
-    true_zone: props.currentRule?.true_zone || false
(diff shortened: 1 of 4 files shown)


src/components/standalone/firewall/rules/FirewallRulesTable.vue
```diff
@@ -97,6 +97,11 @@ const filteredRules = computed(() => {
   }
 })
 
+// enable follows rule.enabled and rule.active_zone
+const enabled = computed(() => {
```
andre8244, line 101: `computed` values usually don't have parameters (`rule` in this case). Computed values with parameters are not [officially documented](https://vuejs.org/guide/essentials/computed.html#computed-properties) and probably don't give any advantages. I suggest to use a simple `isEnabled` method instead

src/i18n/en/translation.json
```diff
@@ -1401,7 +1401,10 @@
       "any_source_address": "Any source address",
       "any_destination_address": "Any destination address",
       "source_object": "Source object",
-      "destination_object": "Destination object"
+      "destination_object": "Destination object",
+      "inactive": "Inactive",
+      "zone_no_longer_exists": "This rule is not active because it uses a zone that no longer exists",
```
andre8244, line 1406:
```suggestion
      "zone_no_longer_exists": "This rule is inactive because it uses a zone that no longer exists",
```

src/i18n/en/translation.json
```diff
@@ -1401,7 +1401,10 @@
...
+      "disabled_rule": "This rule is disabled by its configuration, you can enable it again"
```
andre8244, line 1407:
```suggestion
      "disabled_rule": "This rule has been manually disabled, but can be re-enabled"
```

src/stores/standalone/firewall.ts
```diff
@@ -70,6 +70,7 @@ export interface FirewallRule {
   ns_tag: string[] // tags
   add_to_top: boolean // add rule to top?
   system_rule: boolean // system rule?
+  active_zone: boolean // zone exist?
```
andre8244, line 73:
```suggestion
  active_zone: boolean // zone exists?
```

src/components/standalone/firewall/rules/CreateOrEditFirewallRuleDrawer.vue
```diff
@@ -720,7 +720,8 @@ async function saveRule() {
       return tag.id
     }),
     log: isLoggingEnabled.value,
-    system_rule: props.currentRule?.system_rule || false
+    system_rule: props.currentRule?.system_rule || false,
+    active_zone: props.currentRule?.active_zone || false
```
andre8244, line 724: If I'm not mistaken `active_zone` should not be passed, since it is computed by the backend

stephdl: type script makes me some noise, crying I have a missing property in vscode

stephdl: I recall also the build failed, the previous name of the property was true_zone, I renamed it after active_zone

https://github.com/NethServer/nethsecurity-ui/actions/runs/12596455890/job/35107582027

andre8244 (approved):

PR #482, opened 2025-01-10
fix(port forward): adding tooltip for disabled port forwards

stephdl:
Introduce a tooltip for inactive port forwards when the header contains a hyphen and add translations for the inactive status and range object compatibility.

https://github.com/NethServer/nethsecurity/issues/898

image

image

image

if I revert the host range

image
image

Note:
it seems we need to save the upper unsaved change button if we want the change is effective

### DIFF
--- a/src/i18n/en/translation.json
+++ b/src/i18n/en/translation.json
@@ -1357,7 +1357,7 @@
       "no_object": "No object",
       "port_forwards_for_destination_name": "Port forwards for destination '{name}'",
       "inactive": "Inactive",
-      "range_object_not_compatible": "This port forward is inactive because host set [{name}] include an ip range"
+      "range_object_not_compatible": "This port forward is inactive because host set '{name}' includes an ip range"
     },
     "firewall_rules": {
       "title": "Firewall rules",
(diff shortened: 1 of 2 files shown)


src/i18n/en/translation.json
```diff
@@ -1355,7 +1355,9 @@
       "restricted_object": "Restricted object",
       "restricted_object_tooltip": "All objects supported, except host sets with IP ranges",
       "no_object": "No object",
-      "port_forwards_for_destination_name": "Port forwards for destination '{name}'"
+      "port_forwards_for_destination_name": "Port forwards for destination '{name}'",
+      "inactive": "Inactive",
+      "range_object_not_compatible": "This port forward is inactive because host set '{name}' includes an ip range"
```
gsanchietti, line 1360:
```suggestion
      "range_object_not_compatible": "This port forward is inactive because host set '{name}' includes an IP range"
```

gsanchietti (dismissed):
Just a minor fix on a label.
From a user point of view, it looks good to me

andre8244 (approved):

gsanchietti (approved):

andre8244 (approved):
Remember to squash and use a commit with conventional commit :)

PR #496, opened 2025-01-21
feat(host set): add validation for IP range compatibility with port forwards

stephdl:
Implement validation to ensure that IP ranges are compatible with port forwards, enhancing error messaging and user feedback in the host set management.


src/components/standalone/users_objects/CreateOrEditHostSetDrawer.vue
```diff
@@ -105,7 +106,18 @@ watch(
   }
 )
 
+// compute redirectItemsName the name of the portforward rule using this object
+watch(
+  () => props.currentHostSet?.matches,
+  async () => {
+    if (props.currentHostSet?.matches) {
+      redirectItemsName.value = await getMatchedItemsName(props.currentHostSet.matches)
+    }
```
Tbaile, line 115:
```suggestion
  async (matches) => {
    if (matches) {
      redirectItemsName.value = await getMatchedItemsName(matches)
    }
```

src/components/standalone/users_objects/CreateOrEditHostSetDrawer.vue
```diff
@@ -133,6 +145,36 @@ function runFieldValidators(
   return validators.every((validator) => validator.valid)
 }
 
+async function getMatchedItemsName(matches: string[]): Promise<string> {
+  try {
+    const res = await ubusCall('ns.objects', 'get-info', { ids: matches })
+    const names = []
+    for (const match of Object.values(res.data.info) as any[]) {
```
Tbaile, line 152: Risky call, a Type would be ideal to handle this kind of data
While it's only a type-safe oriented fix (and if API changes at runtime this will fail) let's strengthen the codebase

src/i18n/en/translation.json
```diff
@@ -2203,7 +2203,8 @@
       "host_set_already_exists": "Host set already exists",
       "domain_set_already_exists": "Domain set already exists",
       "delete_domain_set": "Delete domain set",
-      "database_mwan3": "MultiWAN"
+      "database_mwan3": "MultiWAN",
+      "range_not_compatible_with_port_forward": "The ip range is not compatible with an object used in portforward : {name}"
```
andre8244, line 2207:
```suggestion
      "range_not_compatible_with_port_forward": "IP range is not compatible with port forwards. This object is currently used by: {name}"
```

src/components/standalone/users_objects/CreateOrEditHostSetDrawer.vue
```diff
@@ -133,6 +145,36 @@ function runFieldValidators(
...
+      if (match.type === 'redirect') {
+        names.push(match.name)
+      }
+    }
+    return names.join(' ')
```
andre8244, line 157:
```suggestion
    return names.join(', ')
```

src/components/standalone/users_objects/CreateOrEditHostSetDrawer.vue
```diff
@@ -45,6 +45,7 @@ const props = defineProps({
 
 const emit = defineEmits(['close', 'reloadData'])
 
+const redirectItemsName = ref('')
```
andre8244, line 48: Maybe renaming to something like `portForwardsUsingHostSet` is more clear

src/components/standalone/users_objects/CreateOrEditHostSetDrawer.vue
```diff
@@ -105,7 +106,18 @@ watch(
...
+  () => props.currentHostSet?.matches,
+  async () => {
+    if (props.currentHostSet?.matches) {
+      redirectItemsName.value = await getMatchedItemsName(props.currentHostSet.matches)
+    }
+  }
+)
+
 function closeDrawer() {
+  emit('reloadData')
```
andre8244, line 120: Is this needed?

src/components/standalone/users_objects/CreateOrEditHostSetDrawer.vue
```diff
@@ -133,6 +145,36 @@ function runFieldValidators(
...
+      }
+    }
+    return names.join(' ')
+  } catch (error: any) {
+    console.error('Error fetching getMatchedItemsName:', error)
+    return ''
+  }
+}
+
+function noIpRangeWithPortForward(records: Array<string>) {
```
andre8244, line 164: How about renaming to something like `validateNoIpRangeWithPortForward`?

src/components/standalone/users_objects/CreateOrEditHostSetDrawer.vue
```diff
@@ -105,7 +106,18 @@ watch(
   }
 )
 
+// compute redirectItemsName the name of the portforward rule using this object
+watch(
+  () => props.currentHostSet?.matches,
+  async () => {
+    if (props.currentHostSet?.matches) {
```
andre8244, line 113: We need an `else` branch with `redirectItemsName.value = ''`. Try to test this:
- Trigger the validation error about IP range
- Create a new host set containing an IP range
- Even if the new host set is not used by any port forward, the validation error appears again

src/components/standalone/users_objects/CreateOrEditHostSetDrawer.vue
stephdl: yes else the UI keeps the entry you tried to validate when you close the drawer until you reload the whole page

src/components/standalone/users_objects/CreateOrEditHostSetDrawer.vue
stephdl: sorry but vscode is complaining :p

image

src/components/standalone/users_objects/CreateOrEditHostSetDrawer.vue
stephdl: Should be good @Tbaile, is it what you got in mind ?

src/components/standalone/users_objects/CreateOrEditHostSetDrawer.vue
andre8244: Understood. This is a pre-existing glitch, to solve it let's `import { cloneDeep } from 'lodash-es'`  and change line 106 to:
```js
records.value = cloneDeep(props.currentHostSet.ipaddr)
```

src/components/standalone/users_objects/CreateOrEditHostSetDrawer.vue
Tbaile: You're missing the `async (matches)` above! :grimacing:

src/components/standalone/users_objects/CreateOrEditHostSetDrawer.vue
Tbaile: An example of type definitions and all can be found here: https://github.com/NethServer/nethsecurity-ui/blob/f018a2a4d73267899fbd310287b5610aa8c759b1/src/components/standalone/security/ips/IpsEventList.vue#L55-L70

src/components/standalone/users_objects/CreateOrEditHostSetDrawer.vue
stephdl: better, thank

stephdl:
it should be good @andre8244 @Tbaile

andre8244:
It looks like indirect IP ranges are not caught by the validation :cry:

To reproduce:
- Create a host set `myhostset` containing an IP address
- Create a port forward using `myhostset`
- Create a host set `myrange` containing an IP range
- Edit `myhostset` by adding `myrange` to its records
- Save is permitted, but it shouldn't since the port forward cannot be applied on `myrange`
- Below the objects after saving the host set

image

andre8244 (changes requested):
See https://github.com/NethServer/nethsecurity-ui/pull/496#issuecomment-2609511280

src/components/standalone/users_objects/CreateOrEditHostSetDrawer.vue
```diff
@@ -43,8 +44,17 @@ const props = defineProps({
   }
 })
 
+type MatchInfo = {
+  database: string
+  family: string
+  id: string
+  name: string
+  type: string
+}
```
Tbaile, line 53:
```suggestion
type MatchInfo = {
  database: string
  family: 'ipv4' | 'ipv6'
  id: string
  name: string
  type: string
}
```

src/components/standalone/users_objects/CreateOrEditHostSetDrawer.vue
```diff
@@ -43,8 +44,17 @@ const props = defineProps({
...
+
```
Tbaile, line 54:
```suggestion
type MatchInfoResponse = AxiosResponse<{
  info: MatchInfo[]
}>

```

src/components/standalone/users_objects/CreateOrEditHostSetDrawer.vue
```diff
@@ -133,6 +155,36 @@ function runFieldValidators(
   return validators.every((validator) => validator.valid)
 }
 
+async function getMatchedItemsName(matches: string[]): Promise<string> {
+  try {
+    const res = await ubusCall('ns.objects', 'get-info', { ids: matches })
+    const names: string[] = []
+    for (const match of Object.values(res.data.info) as MatchInfo[]) {
+      if (match.type === 'redirect') {
```
Tbaile, line 163:
```suggestion
    const res: MatchInfoResponse = await ubusCall('ns.objects', 'get-info', { ids: matches })
    const names: string[] = []
    for (const match of res.data.info) {
      if (match.type == 'redirect') {
```

Tbaile (changes requested):

src/components/standalone/users_objects/CreateOrEditHostSetDrawer.vue
stephdl: I got 
`
The error "TypeError: res.data.info is not iterable" indicates that res.data.info is not an array or another iterable object. This can happen if the structure of the response data is not as expected.`

I needed to adapt

src/components/standalone/users_objects/CreateOrEditHostSetDrawer.vue
```diff
@@ -133,6 +174,79 @@ function runFieldValidators(
   return validators.every((validator) => validator.valid)
 }
 
+async function getMatchedItemsName(matches: string[]): Promise<string> {
+  try {
+    const res: MatchInfoResponse = await ubusCall('ns.objects', 'get-info', { ids: matches })
+    const names: string[] = []
+    for (const match of Object.values(res.data.info) as MatchInfo[]) {
```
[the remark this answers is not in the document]
stephdl: image

src/components/standalone/users_objects/CreateOrEditHostSetDrawer.vue
```diff
@@ -133,6 +174,79 @@ function runFieldValidators(
...
+
+async function getMatchedObjectName(
+  matches: string[]
+): Promise<{ portForwards: string; objects: string }> {
+  try {
+    const res: MatchInfoResponse = await ubusCall('ns.objects', 'get-info', { ids: matches })
+    const matchedPortForwardNames: string[] = []
+    const objectsUsingPortForward: string[] = []
+    for (const match of Object.values(res.data.info) as MatchInfo[]) {
+      if (match.database === 'objects') {
```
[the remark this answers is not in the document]
stephdl: image

Tbaile (approved):

src/components/standalone/users_objects/CreateOrEditHostSetDrawer.vue
```diff
@@ -86,7 +90,15 @@ const ipVersionOptions = ref([
 ])
 
 const recordOptionsButCurrent = computed(() => {
-  return props.recordOptions?.filter((option) => option.id !== props.currentHostSet?.id)
+  // Filter out objects from recordOptions based on the presence of an IP address with a hyphen in allObjects
+  const objectsWithHyphenIp = props.allObjects
```
andre8244, line 99: This way we are preventing the creation of host sets containing IP ranges, regardless they presence of port forwards. A more correct implementation might be:
- While editing a host set, determine if the current host set is being used by a port forward
- If current host set is used by a port forward, then we need to track which objects contains IP ranges directly or indirectly, by looping `ns.objects get-info` until there are no more object references :confused:
- Finally, we should remove these objects from the Records combobox

andre8244 (changes requested):

src/components/standalone/users_objects/CreateOrEditHostSetDrawer.vue
stephdl: so I based my work on a bug of the UI, actually when you have an object that use another object it is not available to be used as a portforward object

Inevitably searching for a sub-sub-sub-object that has a parent object using port forwarding makes the work much more complicated, searching only for the direct object is simpler.

stephdl:
I removed the validation code of a parent object using port range 

why ?

- in portforward page an object using an hypen (range) is not listed as available in the widget
- an object with another object as record in the object page is not listed as available in the forward page

andre8244 (commented):
Let's fix **Restrict access from** field too

src/components/standalone/users_objects/CreateOrEditHostSetDrawer.vue
```diff
@@ -76,7 +91,15 @@ const ipVersionOptions = ref([
...
+    .filter((obj) => obj.ipaddr.some((ip: string) => ip.includes('-')))
+    .map((obj) => obj.name)
```
andre8244, line 97: Why are we considering `obj.name` instead of `obj.id`? Using the ID should be safer. I guess wel'll need to change line 101 too: `.includes(option.id)` (let's test this if we go this way)

src/i18n/en/translation.json
```diff
@@ -2277,7 +2277,9 @@
       "host_set_already_exists": "Host set already exists",
       "domain_set_already_exists": "Domain set already exists",
       "delete_domain_set": "Delete domain set",
-      "database_mwan3": "MultiWAN"
+      "database_mwan3": "MultiWAN",
+      "range_not_compatible_with_port_forward": "IP range is not compatible with port forwards. This object is currently used by: {name}",
```
andre8244, line 2281:
```suggestion
      "range_not_compatible_with_port_forward": "IP range is not compatible with port forwards. This host set is currently used by: {name}",
```

src/i18n/en/translation.json
```diff
@@ -1364,7 +1364,7 @@
       "enter_restricted_addresses": "Enter restricted addresses",
       "restricted_addresses": "Restricted addresses",
       "restricted_object": "Restricted object",
-      "restricted_object_tooltip": "All objects supported, except host sets with IP ranges",
+      "restricted_object_tooltip": "All objects supported, except host sets with IP ranges or with nested objects",
```
andre8244, line 1367:
```suggestion
      "restricted_object_tooltip": "All objects supported, except host sets containing IP ranges or nested objects",
```

src/components/standalone/users_objects/CreateOrEditHostSetDrawer.vue
```diff
@@ -76,7 +91,14 @@ const ipVersionOptions = ref([
 ])
 
 const recordOptionsButCurrent = computed(() => {
-  return props.recordOptions?.filter((option) => option.id !== props.currentHostSet?.id)
+  // Filter out objects from recordOptions based on the presence of an IP address with a hyphen in allObjects
+  console.log('props.allObjects', props.allObjects)
```
andre8244, line 95: Let's remove this debug print

andre8244 (dismissed):
Let's remove that debug print and then we should be good

Tbaile (approved):

PR #501, opened 2025-01-27
fix(ips): replacing drop with block

Tbaile:
Changes needed due to: https://github.com/NethServer/nethsecurity/pull/1043


andre8244 (approved):

Tbaile:
Holding off this one for the moment, need to check if that's really a needed change (seems to have a discrepancy between configured action and logged one)

PR #508, opened 2025-01-30
fix(threat shield): hide wan zone in ts dns

andre8244:
Do not show WAN zone in Threat shield DNS settings

Ref: 
- https://github.com/NethServer/nethsecurity/issues/906


andre8244:
We are gonna exclude WAN zone in https://github.com/NethServer/nethsecurity/pull/1056

PR #513, opened 2025-02-12
fix(account): added old password input

Tbaile:
Implements
- https://github.com/NethServer/nethsecurity/pull/1077

### DIFF
--- a/src/i18n/en/translation.json
+++ b/src/i18n/en/translation.json
@@ -399,9 +399,7 @@
     "invalid_domain_name": "Invalid domain name",
     "duplicate_rule": "Rule already exists",
     "duplicate_suppression": "Suppression already exists",
-    "key_already_exists": "Key already exists",
-    "key_invalid_format": "Invalid key format",
-    "invalid_old_password": "Old password doesn't match"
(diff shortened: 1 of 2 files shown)


src/i18n/en/translation.json
```diff
@@ -398,7 +398,10 @@
     "cannot_save_blocked_domain": "Cannot save blocked domain",
     "invalid_domain_name": "Invalid domain name",
     "duplicate_rule": "Rule already exists",
-    "duplicate_suppression": "Suppression already exists"
+    "duplicate_suppression": "Suppression already exists",
+    "key_already_exists": "Key already exists",
```
andre8244, line 402: What are this and the following i18n string for?

src/i18n/en/translation.json
```diff
@@ -398,7 +398,10 @@
...
+    "key_invalid_format": "Invalid key format",
+    "invalid_old_password": "Old password doesn't match"
```
andre8244, line 404:
```suggestion
    "invalid_old_password": "Old password is incorrect"
```

andre8244 (approved):

PR #533, opened 2025-04-17
fix(ssh): configure only main dropbear server

gsanchietti:
Make sure to configure the SSH server named main.
Otherwise when the HA is configured, the UI will show and configure dropbear instance dedicated to the HA.

Companion PR: https://github.com/NethServer/nethsecurity/pull/871


Tbaile (approved):
Ok to me, will wait for the uci defaults that changes the default config, do not merge before then

Tbaile (approved):

PR #547, opened 2025-05-08
fix: replace string icons with IconDefinition

andre8244:
Use `IconDefinition` type instead of `string` for `NeRadioSelection` and `NeDropdown` icons.

See:
- https://github.com/nethesis/vue-components/pull/84


Tbaile (approved):
All hail the typing and tree shaking god :pray:

PR #557, opened 2025-05-23
feat: added subscription icon on top

Tbaile:
Lucia pointed out that this was missing, was there a reason why?
image


andre8244:
I guess we never noticed it in the mockups :man_shrugging:

andre8244 (approved):

andre8244 (approved):

PR #563, opened 2025-05-27
feat(nat): added device specification for NATs

Tbaile:
Needs: https://github.com/NethServer/nethsecurity/pull/1235
Closes https://github.com/NethServer/nethsecurity/issues/1183

### DIFF
--- a/src/components/standalone/firewall/nat/CreateOrEditNatRuleDrawer.vue
+++ b/src/components/standalone/firewall/nat/CreateOrEditNatRuleDrawer.vue
@@ -195,7 +195,7 @@ type ListDevicesResponse = AxiosResponse<{
 function listAvailableDevices() {
   loadingAvailableDevices.value = true
   errorAvailableDevices.value = undefined
-  ubusCall('ns.nat', 'list-interfaces')
+  ubusCall('ns.nat', 'list-devices')
     .then((response: ListDevicesResponse) => (availableDevices.value = response.data.devices))
     .catch((reason) => (errorAvailableDevices.value = reason))
     .finally(() => (loadingAvailableDevices.value = false))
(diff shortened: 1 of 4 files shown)


src/components/standalone/firewall/nat/CreateOrEditNatRuleDrawer.vue
```diff
@@ -130,22 +148,26 @@ watch(
       })
       firewallConfig.fetch()
       listHostSuggestions()
+      listAvailableDevices()
       sourceAddress.value = ''
       destinationAddress.value = ''
+      advancedSettingsExpanded.value = false
```
andre8244, line 154: We usually expand advanced settings if the the user made a change to its default settings. We can do this at line 171 instead, according to `props.currentRule.device`

src/components/standalone/firewall/nat/CreateOrEditNatRuleDrawer.vue
```diff
@@ -456,6 +496,33 @@ async function saveRule() {
           :invalid-message="t(errorBag.getFirstI18nKeyFor('snat_ip'))"
           :disabled="loading.saveRule"
         />
+        <NeExpandable
+          :label="t('common.advanced_settings')"
+          :is-expanded="advancedSettingsExpanded"
+          @set-expanded="(ev: boolean) => (advancedSettingsExpanded = ev)"
+        >
+          <NeSkeleton v-if="loadingAvailableDevices" :lines="2" />
+          <NeInlineNotification
```
andre8244, line 505: Maybe it's better to put this error outside the expandable, in order to always show it if the api call fails

src/stores/standalone/firewall.ts
```diff
@@ -81,6 +81,7 @@ export interface NatRule {
   dest_ip: string
   target: NatRuleAction
   snat_ip?: string
+  device: string
```
andre8244, line 84: Since `device` can be an empty string I'd consider it as an optional attribute, but it's just a matter of taste, we can leave it as required

Tbaile: The field is not optional, it's empty, if we check for it's nullability we don't actually check for anything

PR #570, opened 2025-06-03
chore: upgrade tailwindcss to v4

andre8244:
Upgrade TailwindCSS to major version 4

See:
- https://github.com/nethesis/vue-components/pull/86


andre8244:
CI will succeed after https://github.com/nethesis/vue-components/pull/86 is released

PR #594, opened 2025-07-15
fix(port forward): added automated tag if entry is managed by system

Tbaile:
image

Closes: https://github.com/NethServer/nethsecurity/issues/1286

### DIFF
--- a/src/i18n/en/translation.json
+++ b/src/i18n/en/translation.json
@@ -60,7 +60,9 @@
     "confirm": "Confirm",
     "ip_address": "IP address",
     "password_confirmation_required": "Password confirmation required",
-    "password_confirmation_required_description": "Please confirm your password to proceed"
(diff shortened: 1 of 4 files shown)


src/i18n/en/translation.json
```diff
@@ -60,7 +60,9 @@
     "confirm": "Confirm",
     "ip_address": "IP address",
     "password_confirmation_required": "Password confirmation required",
-    "password_confirmation_required_description": "Please confirm your password to proceed"
+    "password_confirmation_required_description": "Please confirm your password to proceed",
+    "automated": "Automated",
+    "automated_tooltip": "This entry is managed by the system and cannot be modified."
```
andre8244, line 65: Not sure this string will be used so often to put it inside `common`. If we want to keep it here, I'd rephrase "This entry" to "This item", otherwise let's move it to port forward section and let's show "This port forward..."

andre8244 (approved):

PR #606, opened 2025-07-30
fix(controller): prevent deletion of used groups

gsanchietti:
See comment https://github.com/NethServer/nethsecurity/issues/1300#issuecomment-3135438983


Tbaile (approved):
Maybe you'll need something in the table to show if the user are assigned to group? No clue how the UI is tho, if they're already there, carry on

PR #746, opened 2026-04-23
feat(migration): move API & UI to new my

edospadoni:
## Summary

Switches the Backup/Restore views off the legacy `backupd` shape and onto the `my`-native JSON produced by `collect`, following the appliance-side cutover in NethServer/nethsecurity#1609.

- Backup listing types + templates now read `uploaded_at` (RFC3339) and `filename`; the epoch `created` / `name` fields are gone.
- `DownloadBackupModal` parses `uploaded_at` and reduces it to an epoch-second integer for the generated file name, dodging the colons that crash some filesystems.
- `DeleteBackupModal` checks the normalized `{message: 'success'}` reply.
- `RunBackupModal` wires `@secondary-click` so the **Close** button actually closes the modal (only the "x" worked before).
- `BackupContent` refreshes the backups store after `SetPassphraseDrawer` succeeds, so the "Passphrase not configured" warning and the disabled Run-backup state flip without a full page reload.
- `RestoreContent`: when the selected backup (or uploaded file) ends in `.gpg`, the **Passphrase** field drops its _Optional_ tag and becomes required with validation focus, since the restore would otherwise fail mid-flight on GPG decrypt.

## Blocking / coordination

> [!WARNING]
> **Do not merge this PR until the appliance-side cutover lands.**
> The new `my`-native JSON shape is produced only once the companion cutover (NethServer/nethsecurity#1609) is deployed on the appliance; on a device still talking to `backupd`, the old fields (`created`, `name`) are missing and the views break.

Requires in order:
1. NethServer/my#81 (`my` backend + collect + UI + security hardening + DELETE route) merged and deployed to production.
2. NethServer/nethsecurity#1608 (dual-send) merged and released.
3. NethServer/nethsecurity#1609 (full cutover of `remote-backup` + `ns.backup` onto `my`) merged and released.
4. This PR merged and released together with the cutover so the appliance UI matches the new payload shape.

Tracking issue: NethServer/my#83.

Subscription page:
image

Hotspot page:
image


edospadoni:
Added the **hotspot login via My Nethesis** (`9209ee9`): primary button running the OIDC device pairing in a popup (SSO with the my.nethesis.it session) with polling until the unit picks the token up; classic credentials form moved under a collapsed "Legacy login" section; connected account/user/manager summary shown above the hotspot configuration; popup-blocker, unsupported-manager and transient-error handling. Pairs with NethServer/nethsecurity#1609 and nethesis/icaro#200.

PR #754, opened 2026-05-05
feat: expose metrics from Victoria

### LINKED ISSUE #1656: Monitoring: adding long time storage for metrics
Firewall needs a long term storage for metrics, at the moment Netdata does the job but it has the following limitations:
- only 6 hours
- very old build
- hard to customize

We landed on [Victoria Metrics](https://victoriametrics.com/) due to the nice space usage and easy replacement for Prometheus, which does the same as Netdata.
However a tool that scrapes the data from the firewall and send it to Victoria is needed, that's where [Telegraf](https://github.com/influxdata/telegraf) comes in.
From Telegraf the data to be sent to the Victoria Metrics is the following:
- generic machine info (CPU, MEM, etc...)
- MWAN status
- services status
- pings

Victoria can through `vmalert` mimic what AlertManager does, which is in firewall's case:
- notify downtimes
- notify alerts

Alerts is kept in storage by Victoria Metrics itself.

gsanchietti:
Since netdata will be removed, switch monitor to Victoria Metrics by including a new page inside Monitoring section.

See companion PR: https://github.com/NethServer/nethsecurity/pull/1633

Example:
image


gsanchietti:
Addressed the review feedback from this round:

- Metrics preferences now use the existing per-user preference helpers and the stored values are validated before use.
- Time-range labels were reduced to parameterized i18n entries.
- TimeLineChart now lets custom options override defaults.
- The empty-state icon now uses faChartLine.
- The Telegraf configuration error uses a dedicated translation key.

All review threads from this pass are resolved.

PR #755, opened 2026-05-06
fix: typo

filippocarletti:


m-dilorenzi:
The use of the term *connection* in the singular form is correct because in the phrase "connection history" it acts as a **noun adjunct**, which is generally used in the singular form with a few specific exceptions.

m-dilorenzi (approved):

PR #763, opened 2026-05-13
fix(ipsec): reject curly braces in pre-shared keys

gsanchietti:
## Summary

Reject unsupported curly brace characters in imported IPsec pre-shared keys.
Add a tooltip next to the pre-shared key field so users know that `{` and `}`
are not accepted.
Cover the new validator with a focused unit test.

## Related issue

NethServer/nethsecurity#1631

## How to test

1. Open the IPsec tunnel drawer and choose a custom pre-shared key.
2. Enter a key containing `{` or `}` and verify the form shows a validation error.
3. Hover the tooltip next to the pre-shared key field and verify it explains
   that `{` and `}` are not accepted.
4. Enter a key without curly braces and verify the form can proceed.

## Dependencies

None


src/i18n/en.json
```diff
@@ -2182,6 +2183,7 @@
       "choose_wan": "Choose WAN",
       "add_network": "Add network",
       "pre_shared_key": "Pre-shared key",
+      "pre_shared_key_invalid_chars_tooltip": "The characters \\{ and \\} are not allowed in the pre-shared key",
```
m-dilorenzi: Should we use `'` char to enclose the curly braces characters to improve the message's readability and clarity?

```suggestion
      "pre_shared_key_invalid_chars_tooltip": "The characters '\\{' and '\\}' are not allowed in the pre-shared key",
```

src/i18n/it.json
```diff
@@ -1563,6 +1564,7 @@
       "wan_ip_address": "Indirizzo IP WAN",
       "enable": "Abilita",
       "pre_shared_key": "Chiave condivisa",
+      "pre_shared_key_invalid_chars_tooltip": "I caratteri { e } non sono consentiti nella chiave condivisa",
```
m-dilorenzi: Same as above. Here is missing the two escape characters
```suggestion
      "pre_shared_key_invalid_chars_tooltip": "I caratteri '\\{' e '\\}' non sono consentiti nella chiave condivisa",
```

src/i18n/ta.json
m-dilorenzi, line 2074: Here is missing the definition of `pre_shared_key_invalid_chars_tooltip` as done for `it` and `en` languages.

m-dilorenzi (changes requested):

src/components/standalone/ipsec_tunnel/CreateOrEditTunnelDrawer.vue
```diff
@@ -606,12 +605,20 @@ watch(
...
         />
 
-        <NeTextInput
-          v-if="presharedKeyMode == 'import'"
-          v-model="presharedKey"
-          :invalid-message="validationErrorBag.getFirstFor('presharedKey')"
-          :label="id ? t('standalone.ipsec_tunnel.pre_shared_key') : ''"
-        />
+        <div v-if="presharedKeyMode == 'import'" class="space-y-2">
+          <NeFormItemLabel class="flex gap-2">
```
andre8244: No need to use `<NeFormItemLabel>` and `<NeTooltip>`, we can use the `label` prop and the `tooltip` slot of `NeTextInput`. Example: https://github.com/NethServer/nethsecurity-ui/blob/main/src/views/standalone/LogsView.vue#L127-L140

src/i18n/ta.json
```diff
@@ -10,6 +10,7 @@
     "required": "தேவை",
     "unique": "தனிப்பட்ட மதிப்பை உள்ளிடவும்",
     "invalid": "புலம் தவறான மதிப்பைக் கொண்டுள்ளது",
+    "curly_braces_not_allowed": "சுருள் அடைப்புக்குறி எழுத்துகள் அனுமதிக்கப்படவில்லை",
```
andre8244: We usually don't touch translation files other than `en` and `it`

Tbaile (approved):
Aside from the escaping of the lang files and the ta lang changed (i have no clue on what to verify if it's actually legit), looking fine

src/i18n/ta.json
gsanchietti: Not needed anymore.

src/i18n/ta.json
gsanchietti: Removed

src/components/standalone/ipsec_tunnel/CreateOrEditTunnelDrawer.vue
gsanchietti: Done

src/i18n/it.json
gsanchietti: Fixed

m-dilorenzi (approved):

PR #768, opened 2026-05-15
feat(ipsec): convert DPD to combobox

### LINKED ISSUE #1298: Invalid configuration: dpd_action = start accepted but not supported by strongSwan
**Problem description:**
In the IPsec tunnel configuration file, when dpd_action is enabled, the dpd_action field is set to set to start, like in the example below:

```
children {
  ns_8b4cbf36_tunnel_1 {
    local_ts = 172.25.0.0/24
    remote_ts = 192.168.1.0/24
    if_id_in = 3
    if_id_out = 3
    start_action = start
    esp_proposals = aes256-sha256-modp2048
    mode = tunnel
    life_time = 66m
    rekey_time = 3600
    dpd_action = start
  }
}

```

However, according to the official [strongSwan swanctl.conf documentation](https://docs.strongswan.org/docs/5.9/swanctl/swanctlConf.html), dpd_action only supports the values:

- clear
- trap
- restart

The value start is not listed and is therefore invalid. This may cause unexpected behavior or mislead users configuring the tunnel.

**Steps to reproduce**
- configure an ipsec tunnel
- enable DPD

**Expected behavior**

In the ipsec configuration `/var/swanctl/swanctl.conf` for each tunnel there will be a line like this:

` dpd_action = restart`

**Actual behavior**

In the ipsec configuration `/var/swanctl/swanctl.conf` for each tunnel there will be a line like this:

` dpd_action = start`

**Suggested fix**
Update the configuration logic to use a valid dpd_action, such as restart, instead of start.

gsanchietti:
Convert DPD radio to NeCombobox with options clear/trap/restart.

Fixes: https://github.com/NethServer/nethsecurity/issues/1298
Related PR (no required for the UI): https://github.com/NethServer/nethsecurity/pull/1676

Screenshot:
image

Available options:
image


Tbaile (changes requested):
Why the need for a lib for a single method that is being used only on the tunnel drawer?

PR #784, opened 2026-06-16
fix: disable alert pooling during updates

Tbaile:
When doing something that distrupts the behind-the-scenes of the firewall (reboot or update) UI errors out since there's a continuos pool for alerts behind the scenes.
This addresses the issue, widens the alert pool and installs tanstack devtools.

### DIFF
--- a/src/components/standalone/update/UpdatePackagesModal.vue
+++ b/src/components/standalone/update/UpdatePackagesModal.vue
@@ -11,6 +11,7 @@ import { NeProgressBar, NeInlineNotification, getAxiosErrorMessage } from '@neth
 import { NeModal } from '@nethesis/vue-components'
 import { ref } from 'vue'
 import { useI18n } from 'vue-i18n'
+import { useSystemActionStore } from '@/stores/standalone/systemAction'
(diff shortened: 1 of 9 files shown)


src/components/standalone/update/UpdatePackagesModal.vue
```diff
@@ -34,6 +36,8 @@ const { startTimer, currentProgress, clearTimer } = useTimer({
   duration: UPDATE_WAIT_TIME,
   progressStep: 0.5,
   onTimerFinish: () => {
+    // Resume polling queries (e.g. alerts): services are back up
+    systemActionStore.setSystemActionInProgress(false)
```
andre8244, line 40: Are we assuming that package updates always complete in less that 20 seconds?

Tbaile: Packages are roughly half MB each, the bigger ones get up to 3 MBs but they are just a handful
With the current implementation of updates, this is the most we can do :shrug:

andre8244 (approved):

PR #793, opened 2026-06-30
feat(threatshield): update geoblocking description for clarity

m-dilorenzi:
This pull request updates the geoblocking feature description in both the English and Italian localization files to clarify that the feature applies specifically to inbound traffic to the firewall, rather than both inbound and outbound traffic.

### DIFF
--- a/src/i18n/en.json
+++ b/src/i18n/en.json
@@ -1831,7 +1831,7 @@
       "asia": "Asia",
       "europe": "Europe",
       "oceania": "Oceania",
-      "geoblocking_description": "Block or allow inbound and outbound traffic based on geographic origin. Select entire continents and define country-level exceptions when needed.",
+      "geoblocking_description": "Block or allow inbound traffic to the firewall based on geographic origin. Select entire continents and define country-level exceptions when needed.",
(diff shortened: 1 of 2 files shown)


Tbaile (approved):
Good for me but I summon @andre8244 that with translations is way more accurate than me

src/i18n/en.json
```diff
@@ -1831,7 +1831,7 @@
       "asia": "Asia",
       "europe": "Europe",
       "oceania": "Oceania",
-      "geoblocking_description": "Block or allow inbound and outbound traffic based on geographic origin. Select entire continents and define country-level exceptions when needed.",
+      "geoblocking_description": "Block or allow inbound traffic to the firewall based on geographic origin. Select entire continents and define country-level exceptions when needed.",
```
andre8244:
```suggestion
      "geoblocking_description": "Block or allow inbound traffic to the firewall unit based on geographic origin. Select entire continents and define country-level exceptions when needed.",
```

andre8244 (approved):
LGTM, I just propose to add "unit" for consistency

PR #796, opened 2026-07-08
perf: dashboard speedup

### LINKED ISSUE #1819: Dashboard: aggregate API calls into a single request for faster loading on older devices
On older/lower-spec devices the dashboard page felt slow to load: the UI polled roughly 17 separate `ns.dashboard` methods (system info, service statuses, counters, VPN tunnels, etc.) on each refresh, each a separate ubus/rpcd round trip.

**Proposed solution**

Add a `dashboard-v2` aggregate method to `ns.dashboard` that returns all dashboard-page data in a single call: `system`, `services`, `counters`, `vpn`, plus placeholders for `threat_shield`, `mac_binding`, `ips`, `ha`. The internet reachability check is intentionally excluded (it pings remote hosts and can take tens of seconds) — it stays on the dedicated `service-status` method.

Degradation contract: if a subsystem call fails, its key is set to `null` and its name is appended to an `errors` array; the rest of the response is still populated, so one failing subsystem doesn't break the whole dashboard.

**Additional context**

Files/packages to change:

- [`packages/ns-api/files/ns.dashboard`](https://github.com/NethServer/nethsecurity/blob/main/packages/ns-api/files/ns.dashboard) — new aggregate method
- `ns-ui` (nethsecurity-ui) — dashboard page must call the new aggregate endpoint instead of the per-card calls

Tbaile:
Added dashboard helpers for new request format
Refactored cards with the dashboard helpers

While there:
    - moved from font-awesome-icon to FontAwesomeIcon
    - migrated to NeBadgeV2
    - using imported icons where possible

Needs: https://github.com/NethServer/nethsecurity/pull/1774

Closes NethServer/nethsecurity#1819


Tbaile:
Holding off this for the moment since I've found a possible better way of handling it.

andre8244:
> Holding off this for the moment since I've found a possible better way of handling it.

Ok, you can ask again for review if/when ready

andre8244 (approved):

PR #798, opened 2026-07-13
fix(ui): preserve page size for lease tables

andre8244:
Toggling loading via `v-if/v-else` unmounted LeasesTable on every
refetch, resetting its selected page size back to the default.
Switch to `v-show` so the skeleton overlays the table instead of
replacing it.


Tbaile (approved):
🤷

andre8244:
> 🤷

Is something unclear? The fix is quite trivial

Tbaile:
What I meant is: if it works :shrug:

PR #815, opened 2026-09-08
feat(dpi): refactor DPI UI

m-dilorenzi:
This PR contains a rework of the DPI Filter UI, replacing the old card-based rule editor with a table-driven interface and introducing application groups as a first-class concept.

### Main changes

- **DPI rules**: rules are now shown in a sortable table with drag & drop reordering, inline enable/disable, and dedicated create/edit and rename drawers replacing the previous multi-purpose modal.
- **Application groups**: new tab (replacing *Exceptions*) to create, edit and delete reusable groups of applications, protocols and their categories, with a guided multi-step creation flow, search and pagination.
- **Data layer**: DPI logic moved out of components into dedicated composables (`useDpiRules`, `useApplicationGroups`, `useDpiCatalog`) built on TanStack  Query, with typed payloads, shared query keys and centralized cache invalidation. The catalog composable merges the netifyd catalog with the loaded applications/protocols and exposes categories with icons and availability info.
- **Subscription awareness**: loaded application/protocol queries are invalidated when the subscription status changes, so the catalog reflects what is actually available.
- **Shared components**: `NeMultiTextInput` gains a `helperText` prop and improved spacing; `NeStepper` gains configurable bar size and color classes.
- **Tests**: added unit tests for the three new composables.
- **i18n**: new English and Italian strings for rules and application groups.


src/stores/standalone/subscription.ts
```diff
@@ -23,7 +29,14 @@ export const useSubscriptionStore = defineStore('subscription', () => {
...
       .then((res: SubscriptionStatusResponse) => {
-        isActive.value = res.data.active ?? false
+        const active = res.data.active ?? false
+
+        if (active !== isActive.value) {
+          queryClient.invalidateQueries({ queryKey: DPI_LOADED_APPLICATIONS_KEY })
+          queryClient.invalidateQueries({ queryKey: DPI_LOADED_PROTOCOLS_KEY })
+        }
+
+        isActive.value = active
```
Tbaile: We can just invalidate the queries, it's data saved in the machine, not a big issue

src/components/standalone/NeStepper.vue
```diff
@@ -1,25 +1,40 @@
...
+  currentStep,
+  stepLabel,
+  barSize = 'sm',
+  labelColorClasses = 'text-primary-600 dark:text-primary-500',
+  barColorClasses = 'bg-primary-600 dark:bg-primary-500'
+} = defineProps<{
   totalSteps: number
   currentStep: number
   stepLabel: string
+  barSize?: 'sm' | 'md' | 'lg' | 'xl'
```
Tbaile: Let's keep just what we use
```suggestion
  barSize?: 'sm' | 'md'
```

Tbaile: Why generating more than we need?

src/components/standalone/StandaloneTopBar.vue
Tbaile, line 1: Unrelated fix, let's not change what we don't need

src/composables/useApplicationGroups.ts
```diff
@@ -0,0 +1,189 @@
...
+          search: params.search.value || undefined,
+          limit: params.pageSize.value,
+          page: params.page.value
+        },
+        { signal }
+      ),
+    select: (res) => res.data.values,
+    placeholderData: keepPreviousData
+  })
+}
```
Tbaile: Not liking this separation, should be one single call with optional parameters

src/composables/useApplicationGroups.ts
```diff
@@ -0,0 +1,189 @@
...
+
+export function useCreateApplicationGroup() {
+  const invalidate = useApplicationGroupsInvalidation()
+
+  return useMutation({
+    mutationFn: (payload: DpiApplicationGroupPayload) =>
+      ubusCall<AppGroupIdResponse>('ns.dpi', 'add-appgroup', payload),
+    onSuccess: invalidate
+  })
+}
```
Tbaile: This is a single usage function inside a separate file, should stay in the create modal

src/composables/useApplicationGroups.ts
```diff
@@ -0,0 +1,189 @@
...
+
+export function useEditApplicationGroup() {
+  const invalidate = useApplicationGroupsInvalidation()
+
+  return useMutation({
+    mutationFn: (payload: DpiApplicationGroupPayload & { id: string }) =>
+      ubusCall<AppGroupIdResponse>('ns.dpi', 'edit-appgroup', payload),
+    onSuccess: invalidate
+  })
+}
```
Tbaile: This is a single usage function inside a separate file, should stay in the edit drawer

src/composables/useApplicationGroups.ts
```diff
@@ -0,0 +1,189 @@
...
+
+export function useDeleteApplicationGroup() {
+  const invalidate = useApplicationGroupsInvalidation()
+
+  return useMutation({
+    mutationFn: (id: string) =>
+      ubusCall<{ data: { message: string } }>('ns.dpi', 'delete-appgroup', { id }),
+    onSettled: invalidate
+  })
+}
```
Tbaile: This is a single usage function inside a separate file, should stay in the delete modal

src/composables/useApplicationGroups.ts
Tbaile, line 1: Most of the functions here are separated for no actual reason, should stay in the Vue file that owns the data, since we're using Tanstack the data is saved into session, so we can avoid repeating it

src/composables/useDpiCatalog.ts
Tbaile, line 1: This file has the same issue, having separate exported functions that are being used just one time

src/components/standalone/dpi/RenameDpiRuleDrawer.vue
```diff
@@ -0,0 +1,109 @@
...
+const { t } = useI18n()
+
+const name = ref('')
+const validationBag = ref(new MessageBag())
+
+const { mutate: renameRule, isPending } = useRenameDpiRule()
+
+const schema = v.object({
+  name: v.pipe(v.string(), v.trim(), v.minLength(1, 'required'), v.maxLength(64, 'name_too_long'))
+})
```
Tbaile: Let validation be done in the backend, single point of truth

src/composables/useDpiCatalog.ts
```diff
@@ -0,0 +1,374 @@
...
+  return useQuery({
+    queryKey: DPI_LOADED_PROTOCOLS_KEY,
+    queryFn: ({ signal }) =>
+      ubusCall<Response<LoadedEntry[]>>('ns.dpi', 'list-loaded-protocols', {}, { signal }),
+    select: (res) => res.data.values,
+    staleTime: 5 * 60 * 1000
+  })
+}
+
+export function useDpiCatalog(kind: Ref<DpiCatalogKind>, categoryId: Ref<string>) {
```
Tbaile, line 171: Why not making this into the backend? Is there a reason? A single call for all the data joined

src/components/standalone/NeMultiTextInput.vue
```diff
@@ -51,6 +52,18 @@ const emit = defineEmits(['delete-item', 'add-item', 'update:modelValue'])
 const keys = ref<string[]>([])
 const items = ref<string[]>([])
 
+const showHelperText = computed(
+  () =>
+    !!props.helperText &&
```
Tbaile: This is to check if the text is true or false, you want an `undefined` check

src/components/standalone/dpi/DeleteDpiRuleModal.vue
```diff
@@ -1,99 +1,74 @@
...
+  close: []
+  deleted: []
+}>()
 
 const { t } = useI18n()
 
-const loading = ref({
-  deleteRule: false
-})
+const deleteError = ref<Error>()
```
Tbaile: this can be extracted by error inside the mutation

src/components/standalone/dpi/CreateApplicationGroupModal.vue
```diff
@@ -0,0 +1,716 @@
...
+
+const nameSchema = v.object({
+  name: v.pipe(
+    v.string(),
+    v.trim(),
+    v.minLength(1, 'required'),
+    v.maxLength(NAME_MAX_LENGTH, 'name_too_long')
+  )
+})
+
```
Tbaile: Move this into backend

src/components/standalone/dpi/CreateApplicationGroupModal.vue
Tbaile, line 1: This approach might not be the best. While all code seem fine we are having a single component behave completely differently by it's state, probably extracting a presentation component and then two components that apply logic (create/edit) is the best approach here

src/components/standalone/dpi/DeleteApplicationGroupModal.vue
```diff
@@ -0,0 +1,98 @@
...
+}>()
+
+const emit = defineEmits<{
+  close: []
+  deleted: []
+}>()
+
+const { t } = useI18n()
+
+const deleteError = ref<Error>()
```
Tbaile: Error can be extracted by the mutation, if you need different behaviours you can just apply a watch function

src/components/standalone/dpi/CreateOrEditDpiRuleDrawer.vue
Tbaile, line 1: This might have the same "issue" of the CreateApplicationModal, single component for multiple usages

Tbaile (commented):
The notes I've left in the useApplicationGroups are valid for all the new composables, also there's no point to test looping functions, tests only things that eventually will break if data underneath changes or edge cases.
Validation should stick to the backend *unless needed*, the only validation I could get behind is the step-by-step one, and I'm starting to think that is better to hit a backend API for validation only.

PR #817, opened 2026-09-15
Threat shield ip labels

cotosso:
Change labels related to logs of blocked packets, in order to clarify the real behavior.


cotosso:
modifications already done in the main banIP PR

